PR #1028, opened 2026-04-29
fix(mahjong): stop background music on unmount (#1020)

### LINKED ISSUE #1020: Mahjong: Background music continues playing after leaving the game
## Summary

Mahjong's background music does not stop when the player navigates away from the game. The music carries over into the lobby and other games.

## Steps to Reproduce

1. Enter Mahjong with music playing (not muted).
2. Leave the game and return to the lobby.
3. Music continues playing in the lobby and persists into other games.

## Additional Symptom

Once the music has leaked outside the game, the mute button no longer silences it — suggesting the leaked audio instance is disconnected from the mute control. Within the game itself, the mute button works correctly.

## Expected Behavior

Background music should stop as soon as the player leaves Mahjong, the same way Starswarm correctly stops its music on exit.

## Proposed Fix

Ensure the Mahjong music track is stopped (and its audio instance cleaned up) when the component unmounts / the player navigates away from the game screen.

## Reference

Starswarm handles this correctly and can serve as the reference implementation.

wcmchenry3-stack:
## Summary

- `expo-audio`'s `AudioPlayer.remove()` frees the JS wrapper but does **not** halt native playback, causing Mahjong's background music to leak into the lobby and other games when the player navigates away mid-game.
- Added an explicit `pause()` call before `remove()` in `useBackgroundMusic`'s unmount cleanup so audio stops immediately on screen exit.
- Added `pause` to the global `expo-audio` Jest mock in `jest.setup.ts` so all screen tests pass cleanly.
- Added a dedicated `useBackgroundMusic.test.ts` with 6 tests, including two that assert `pause()` is called before `remove()` on unmount.

## Root cause

Starswarm "worked" by accident — players normally reach game-over before navigating away, which sets `bgMusicActive=false` and pauses the player before unmount. Mahjong players frequently leave mid-game, so the player was still active when `remove()` was called without a prior `pause()`.

## Files changed

- `frontend/src/game/_shared/useBackgroundMusic.ts` — pause before remove in unmount cleanup
- `frontend/src/game/_shared/__tests__/useBackgroundMusic.test.ts` — new test suite (6 tests)
- `frontend/jest.setup.ts` — add `pause` to global expo-audio mock

## Test plan

- [ ] Run `npx jest --no-coverage --testPathPattern="useBackgroundMusic"` — all 6 pass
- [ ] Run `npx jest --no-coverage --testPathPattern="MahjongScreen"` — all tests pass
- [ ] Manual: enter Mahjong with music on, navigate back to lobby — music stops

Closes #1020

### DIFF
--- a/frontend/jest.setup.ts
+++ b/frontend/jest.setup.ts
@@ -64,6 +64,7 @@ jest.mock("react-native-reanimated", () => {
 jest.mock("expo-audio", () => ({
   createAudioPlayer: jest.fn(() => ({
     play: jest.fn(),
+    pause: jest.fn(),
     seekTo: jest.fn(),
     remove: jest.fn(),
   })),
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -19,7 +19,7 @@
         "@react-navigation/native-stack": "^7.14.6",
         "@sentry/react-native": "~7.11.0",
         "@shopify/react-native-skia": "2.4.18",
-        "expo": "~55.0.17",
+        "expo": "~55.0.18",
         "expo-audio": "~55.0.14",
         "expo-blur": "~55.0.14",
         "expo-haptics": "~55.0.14",
@@ -8642,9 +8642,9 @@
       }
     },
     "node_modules/expo": {
-      "version": "55.0.17",
-      "resolved": "https://registry.npmjs.org/expo/-/expo-55.0.17.tgz",
-      "integrity": "sha512-yVF2phiPw5XgOCedC/oQaL3j0XbwzsBLst3JiAF8bi9aFlxLOVvuDEM8BDg3E09XGSLaGCAclY4q5L+sFerXlQ==",
+      "version": "55.0.18",
+      "resolved": "https://registry.npmjs.org/expo/-/expo-55.0.18.tgz",
+      "integrity": "sha512-J3LVgN8ygERC0pmSjXfW2W/jlT18+VBek6vB9DBJiCNyrGKpSE4Kv9BH7VooiIMEizwwzsgDgXbDRWBS14IaKA==",
       "license": "MIT",
       "dependencies": {
         "@babel/runtime": "^7.20.0",
@@ -8665,7 +8665,7 @@
         "expo-file-system": "~55.0.17",
         "expo-font": "~55.0.6",
         "expo-keep-awake": "~55.0.6",
-        "expo-modules-autolinking": "55.0.18",
+        "expo-modules-autolinking": "55.0.19",
         "expo-modules-core": "55.0.23",
         "pretty-format": "^29.7.0",
         "react-refresh": "^0.14.2",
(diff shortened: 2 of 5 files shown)


wcmchenry3-stack (commented):
## Code Review — PR #1028: `fix(mahjong): stop background music on unmount`

### Overview

Fixes a music-leaking bug where `expo-audio`'s `AudioPlayer.remove()` doesn't halt native playback on all platforms. The fix adds an explicit `pause()` call before `remove()` in the unmount cleanup of `useBackgroundMusic`. Accompanies the fix with a new 6-test suite and fills a missing `pause` mock in the global Jest setup.

---

### The Fix (`useBackgroundMusic.ts`)

**Correct and minimal.** The null-guard before the two-call sequence is the right approach since optional chaining can't be used across two separate calls on the same ref without a double dereference. The comment explains a non-obvious platform behavior — appropriate per project standards.

No issues here.

---

### Tests (`useBackgroundMusic.test.ts`)

**Generally strong, one weak test.**

- The `pauseOrder` array pattern for asserting call ordering is clean and readable.
- The "pause before remove" test for the `active: true → false → unmount` path is valuable — it explicitly pins down `["pause", "pause", "remove"]`, which documents that the active-transition effect and the unmount effect both fire `pause`.

**Issues:**

1. **Weak mute test** — the "mute" describe block doesn't actually test mute toggling. It clears `mockPause`, then immediately checks `typeof mockPause === "function"` (always true), then rerenders with `active: false` and asserts `pause` was called — which is just re-testing the `active` flag path, not the mute path. Either test real mute behavior via `SoundContext` or remove this block. As written, it provides false confidence.

2. **`SOUND_REGISTRY` mutation in `beforeEach`** — deleting keys from the live module-level registry (`delete SOUND_REGISTRY[k]`) is fragile. If the registry gains non-deletable properties (e.g., via `Object.freeze` or a getter), this silently fails. Consider instead mocking the module or making the test-specific key additive with an `afterEach` cleanup.

3. **Missing `seekTo` in local mock** — the local `jest.mock("expo-audio", ...)` omits `seekTo` which the global mock includes. Not a bug today, but if the hook calls `seekTo` in a future code path, tests here would throw rather than fail clearly.

---

### `jest.setup.ts`

Adding `pause` to the global mock is the right fix. No issues.

---

### Risks / Edge Cases

- **Double `pause()` on unmount** when the `active` prop is already `false` — the second test pins this as `["pause", "pause", "remove"]`, so it's documented behavior, and double-pausing audio is benign. Fine to leave as-is, but worth noting that the unmount always pauses regardless of current state (intentional defensive behavior).
- **Starswarm** — the PR description claims Starswarm was unaffected because players typically reach game-over before navigating away. Worth noting that `useBackgroundMusic` is shared, so this `pause`-before-`remove` fix applies there too (harmless, but worth awareness if Starswarm ever gets automated music-stop tests).

---

### Summary

| Area | Verdict |
|---|---|
| Bug fix correctness | Correct |
| Code quality | Clean |
| Comment | Appropriate (non-obvious platform behavior) |
| Test coverage | Good, with one weak test |
| Global mock update | Correct |

**Recommend merging after fixing or removing the mute test** — it asserts nothing meaningful and inflates apparent test coverage. Everything else is solid.

PR #1104, opened 2026-04-30
feat(mahjong): lock to landscape for ~2× tile size on iPhone

### LINKED ISSUE #1103: [Mahjong] Tiles are too small on iPhone — lock to landscape for 2× size improvement
### Why tiles are tiny today

The game board has a fixed natural width of **572px** (`BOARD_W`, `GameCanvas.tsx:36` / `GameCanvas.web.tsx:33`), built from:

```
BOARD_W = PAD_X(10) + 12 columns × TILE_W(44) + 4 layers × LAYER_DX(6) + PAD_X(10) = 572px
```

`MahjongScreen.tsx:486` scales the whole board down to fit the screen:

```typescript
const MAX_TILE_W = 72;
const scale = outerWidth > 0 ? Math.min(MAX_TILE_W / TILE_W, outerWidth / BOARD_W) : 1;
```

On an **iPhone 14 portrait** (390px wide, ~366px after padding):
- `scale = Math.min(72/44, 366/572) = Math.min(1.636, 0.640) = **0.640**`
- Effective tile size: `44 × 0.64 = **28px wide, 36px tall**`
- The `MAX_TILE_W = 72` cap **never applies on phones** — the board is already too wide to upscale in portrait

28×36px tiles are postage-stamp sized. You have to strain to read the face and tap precisely to hit them.

### What landscape gives us for free

On **iPhone 14 landscape** (844px wide, ~804px after safe area):
- `scale = Math.min(1.636, 804/572) = Math.min(1.636, **1.406**) = **1.406**`
- Effective tile size: `44 × 1.406 = **62px wide, 79px tall**`
- That is **2.2× bigger** than portrait — no changes to drawing code required

The turtle layout is naturally wider than tall (572×508), making landscape the correct orientation for this game. Holding the phone sideways for Mahjong is standard and expected.

### 1. Lock Mahjong screen to landscape

Use `expo-screen-orientation` (already a project dependency) to force landscape when the Mahjong screen mounts, and restore on unmount:

```typescript
// MahjongScreen.tsx — on mount/unmount
import * as ScreenOrientation from "expo-screen-orientation";

useFocusEffect(useCallback(() => {
  ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.LANDSCAPE);
  return () => ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.DEFAULT);
}, []));
```

### 2. Raise `MAX_TILE_W` for tablets

The current cap of `72px` (`MahjongScreen.tsx:485`) limits tile size on iPads. Raise it to `100px` so tablets also benefit from the larger canvas:

```typescript
const MAX_TILE_W = 100; // was 72
```

On an iPad Pro 12.9" landscape (~1300px usable): `scale = min(100/44=2.27, 1300/572=2.27)` → tiles would be ~100px wide.

### 3. Verify `outerWidth` accounts for landscape safe-area insets

iPhone models with a notch/Dynamic Island have horizontal safe-area insets in landscape. Confirm the `onLayout` measurement at `MahjongScreen.tsx:488` captures the correct inner width after `GameShell` applies its `paddingLeft`/`paddingRight` (currently `Math.max(insets.left, 12)`). If the board overflows on notched iPhones, increase the horizontal padding floor in landscape.

## Files to Change

- `frontend/src/screens/MahjongScreen.tsx` — orientation lock + raise `MAX_TILE_W`
- No changes needed to `GameCanvas.tsx` or `GameCanvas.web.tsx` — the scale transform handles everything

## Acceptance Criteria

- [ ] On iPhone (any model), opening Mahjong automatically rotates to landscape
- [ ] Leaving Mahjong restores the app's default portrait orientation
- [ ] Effective tile width is ≥ 60px on an iPhone 14 (2× improvement over current ~28px)
- [ ] Board does not overflow or clip on notched/Dynamic Island iPhones in landscape
- [ ] iPad renders tiles at ≤ 100px (board fills screen without overshooting)
- [ ] Existing scroll/shake/flash animations still work correctly at the new scale

wcmchenry3-stack:
$(cat <<'EOF'
## Summary

- **Root cause**: The Mahjong board is 572px wide but iPhones in portrait are ~366px usable, forcing a scale of ~0.64 — tiles render at ~28×36px, too small to read or tap comfortably
- **Fix**: Lock the Mahjong screen to landscape on focus (phone is ~804px usable), which pushes scale to ~1.41 — tiles hit ~62×79px (~2.2× bigger) with zero drawing-code changes
- **Restores portrait** automatically when the user leaves the screen

## Changes

| File | Change |
|---|---|
| `frontend/app.json` | `orientation: "portrait"` → `"default"` — required so iOS Info.plist includes landscape orientations, enabling the per-screen API to work |
| `frontend/package.json` | Add `expo-screen-orientation ~55.0.13` |
| `frontend/src/screens/MahjongScreen.tsx` | `useFocusEffect` locks to `LANDSCAPE` on focus, restores `PORTRAIT_UP` on blur; skipped on web; `MAX_TILE_W` raised 72 → 100 for tablet benefit |

## Test plan

- [ ] Open Mahjong on iPhone — device rotates to landscape automatically
- [ ] Tiles are visibly larger and easier to read/tap
- [ ] Navigate back — device returns to portrait
- [ ] All other game screens remain portrait-only
- [ ] No board overflow or clipping on notched/Dynamic Island iPhones in landscape
- [ ] Web build unaffected (orientation lock skipped via `Platform.OS !== "web"` guard)

Closes #1103

https://claude.ai/code/session_011cNwaZNCYbFjPvPtCdTuVy
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_011cNwaZNCYbFjPvPtCdTuVy)_

### DIFF
--- a/frontend/app.json
+++ b/frontend/app.json
@@ -3,7 +3,7 @@
     "name": "BC Arcade",
     "slug": "gaming-app",
     "version": "1.0.9",
-    "orientation": "portrait",
+    "orientation": "default",
     "icon": "./assets/icon.png",
     "userInterfaceStyle": "light",
     "splash": {
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -26,6 +26,7 @@
         "expo-linear-gradient": "~55.0.13",
         "expo-localization": "^55.0.13",
         "expo-modules-core": "^55.0.22",
+        "expo-screen-orientation": "~55.0.13",
         "expo-status-bar": "^55.0.5",
         "i18next": "^25.10.5",
         "i18next-resources-to-backend": "^1.2.1",
@@ -8845,6 +8846,16 @@
         "react-native-worklets": {
           "optional": true
         }
+      }
+    },
+    "node_modules/expo-screen-orientation": {
+      "version": "55.0.13",
+      "resolved": "https://registry.npmjs.org/expo-screen-orientation/-/expo-screen-orientation-55.0.13.tgz",
+      "integrity": "sha512-nIqZdaUGBQpmJP5A5Te1AxIR5chKLICj5he6EjOzqae00WC4JPadlC+thWCngvmAW6cvINUahd9+leOhLQ2q4w==",
+      "license": "MIT",
+      "peerDependencies": {
+        "expo": "*",
+        "react-native": "*"
       }
     },
     "node_modules/expo-server": {
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -65,6 +65,7 @@
     "expo-linear-gradient": "~55.0.13",
     "expo-localization": "^55.0.13",
     "expo-modules-core": "^55.0.22",
+    "expo-screen-orientation": "~55.0.13",
     "expo-status-bar": "^55.0.5",
     "i18next": "^25.10.5",
     "i18next-resources-to-backend": "^1.2.1",
--- a/frontend/src/components/mahjong/GameCanvas.web.tsx
+++ b/frontend/src/components/mahjong/GameCanvas.web.tsx
@@ -271,8 +271,12 @@ export default function GameCanvas({ state, onTilePress, onShufflePress, onNewGa
       const canvas = canvasRef.current;
       if (!canvas) return;
       const rect = canvas.getBoundingClientRect();
-      const tapX = e.clientX - rect.left;
-      const tapY = e.clientY - rect.top;
+      // getBoundingClientRect reflects the parent's CSS scale transform, so
+      // divide by the visual/native ratio to get canvas drawing coordinates.
+      const scaleX = rect.width / BOARD_W;
+      const scaleY = rect.height / BOARD_H;
+      const tapX = (e.clientX - rect.left) / scaleX;
+      const tapY = (e.clientY - rect.top) / scaleY;
       const tileId = hitTest(state.tiles, tapX, tapY);
       if (tileId !== null) onTilePress(tileId);
     },
--- a/frontend/src/screens/MahjongScreen.tsx
+++ b/frontend/src/screens/MahjongScreen.tsx
@@ -41,8 +41,9 @@ import Animated, {
 } from "react-native-reanimated";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
 import { useTranslation } from "react-i18next";
-import { useNavigation } from "@react-navigation/native";
+import { useFocusEffect, useNavigation } from "@react-navigation/native";
 import { NativeStackNavigationProp } from "@react-navigation/native-stack";
+import * as ScreenOrientation from "expo-screen-orientation";
 
 import { HomeStackParamList } from "../../App";
 import { useTheme } from "../theme/ThemeContext";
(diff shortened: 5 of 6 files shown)


wcmchenry3-stack:
## E2E Debugging Summary (session 2)

### What was done

1. **Removed top-level `expo-screen-orientation` import** (`51b04926`) — the import was included in the web bundle and caused a crash. Replaced with inline `require()` inside a `Platform.OS !== 'web'` guard so the native module is never evaluated in the browser.

2. **Fixed click coordinate translation in `GameCanvas.web.tsx`** (`0717596e`) — the canvas sits inside a parent `View` with `transform: [{ scale }]` (CSS transform). `getBoundingClientRect()` returns the visual (post-transform) bounding box, so we must divide the raw viewport offset by the scale ratio to get canvas drawing coordinates:
   ```ts
   const scaleX = rect.width / BOARD_W;
   const tapX = (e.clientX - rect.left) / scaleX;
   ```

Both fixes are committed and CI confirmed deployed. The Playwright E2E test still fails.

---

### Remaining failure

**Test**: `mahjong-smoke.spec.ts:56:7 › interacting with the board does not crash the app`  
**Error**: `element(s) not found` — the `<canvas role="img" aria-label="Mahjong Solitaire…">` element is completely absent from the DOM after the click (not just hidden, but gone).

Since the canvas is only rendered when `state !== null`, disappearing = the component unmounted or state was set to null. No code path in `handleTilePress`, `handleShuffle`, `handleUndo`, or any `useEffect` explicitly sets state to null after initialization.

**Key difference vs dev** (where the test passes):  
- `dev`: `MAX_TILE_W = 72` → `scale = 1.636` → canvas 936 × 831 px (does not fill viewport width)  
- This PR: `MAX_TILE_W = 100` → `scale = 2.196` → canvas **1256 × 1115 px** (fills the full 1256 px content area)

The scale is constrained by `Math.min(MAX_TILE_W / TILE_W, outerWidth / BOARD_W)`. With MAX_TILE_W=100 on a 1280 px CI viewport the outerWidth becomes the binding constraint; with MAX_TILE_W=72 the tile-size cap is binding.

---

### Recommended next steps

1. **Quick validation**: revert `MAX_TILE_W` from 100 → 72 in `MahjongScreen.tsx:500`, push, and watch CI. If the test turns green, confirm MAX_TILE_W is the root cause.

2. **If confirmed**: investigate what changes between scale=1.636 and scale=2.196 that would cause a navigation or unmount. Candidates:
   - A React error thrown during `Reanimated` worklet execution when the transform value is very large (> 2.0), if there's a Reanimated invariant check.
   - An uncaught exception in `ensureSyncStarted` / `syncStart` that unwinds the component tree (error boundary unmount). Add a try/catch around `syncStart` in `handleTilePress` to test.
   - A scroll event triggered by the click in the ScrollView when the content is very tall (1115 px vs 831 px), causing a layout recalculation that briefly sets `outerWidth = 0` → `scale = 1` → re-render loop.

3. **Alternative**: keep MAX_TILE_W=100 but add a `min()` cap so `boardWrap height ≤ viewport height`. Something like: `const scale = outerWidth > 0 ? Math.min(MAX_TILE_W / TILE_W, outerWidth / BOARD_W, outerHeight / BOARD_H) : 1;` — keeping the board within the visible area prevents any scroll/overflow interaction with the test.

Issue #1458, opened 2026-05-10
Star Swarm: boost Boss firepower so it's clearly more dangerous than Elites


wcmchenry3-stack:
## Problem

The Boss has 4 HP vs the Elite's 2 HP, but its effective firepower is only marginally higher. Both fire precision-aimed bullets dealing 1 damage. The Boss fires bursts of 2–3 shots then goes quiet for 3–4 seconds — a gap wide enough that it doesn't feel more threatening than an Elite in active combat.

The design intent is that players should feel genuine pressure to knock Bosses out early. Right now there's not enough incentive because when a Boss *does* start attacking, it isn't noticeably scarier than an Elite.

## Proposed Changes

All constants live in `frontend/src/game/starswarm/engine.ts`.

| Parameter | Current | Proposed | Effect |
|---|---|---|---|
| Burst size | 2–3 shots | 3–5 shots | Longer volleys feel visually and mechanically heavier |
| `BURST_PAUSE_BASE` | 3000 ms | 2000 ms | Tighter cycle — Boss keeps sustained pressure instead of going silent |
| Boss bullet speed | same as Elite | slightly faster | Same precision aim, less time to dodge — qualitative "boss-tier" feel |

Bullet damage stays at 1 to keep it fair on lower difficulty tiers.

## Acceptance Criteria

- [ ] Boss burst fires 3–5 shots (random) per cycle
- [ ] Pause between bursts is ~2000 ms base (+ existing jitter)
- [ ] Boss bullets travel noticeably faster than Elite bullets
- [ ] Existing `engine.test.ts` passes
- [ ] Feels meaningfully more dangerous than an Elite at all difficulty tiers

wcmchenry3-stack:
Resolved by #1460 (merged).

PR #1460, opened 2026-05-10
feat(starswarm): boost Boss firepower — bigger bursts, faster bullets, shorter pause (#1458)

wcmchenry3-stack:
Boss burst size raised from 2–3 to 3–5 shots; BURST_PAUSE_BASE dropped from
3000 ms to 2000 ms so bursts cycle faster; Boss bullets now travel at 0.46 px/ms
(vs Elite's 0.35) across formation, dive, and circle phases. aimVelocity gains
an optional speed param to support the faster boss bullet without touching Elite
or Grunt behaviour.

https://claude.ai/code/session_01V81eUam5bZoXudJSv2c8Zw

### DIFF
--- a/frontend/src/game/starswarm/__tests__/engine.test.ts
+++ b/frontend/src/game/starswarm/__tests__/engine.test.ts
@@ -1335,12 +1335,8 @@ describe("Boss burst-fire (#979)", () => {
       }),
     };
     s = tick(s, 16, NO_INPUT);
-    const bossBullet = s.enemyBullets.find(
-      (b) => b.vy === BOSS_BULLET_VY
-    );
-    const eliteBullet = s.enemyBullets.find(
-      (b) => b.vy === BULLET_E_VY
-    );
+    const bossBullet = s.enemyBullets.find((b) => b.vy === BOSS_BULLET_VY);
+    const eliteBullet = s.enemyBullets.find((b) => b.vy === BULLET_E_VY);
     expect(bossBullet).toBeDefined();
     expect(eliteBullet).toBeDefined();
   });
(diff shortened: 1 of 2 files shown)


wcmchenry3-stack (commented):
## Code Review

### Code Quality

**`aimVelocity` refactor is correct and clean.**
The vx cap was updated to `Math.max(-speed, Math.min(speed, rawVx))`, which preserves the ≤45° from-vertical constraint correctly for any speed value. The default parameter keeps existing call sites unchanged.

**Burst formula is clearer now.**
Old: `2 + Math.floor(rng() * 2) - 1` → yields 1 or 2 remaining shots (equivalent to `1 + Math.floor(rng() * 2)`, which was needlessly convoluted). New: `2 + Math.floor(rng() * 3)` → yields 2, 3, or 4 remaining shots (3–5 total burst). Clean and easy to read.

**`tickCircling` tier check is slightly ad-hoc.**

```ts
const speed = enemy.tier === "Boss" ? BOSS_BULLET_VY : undefined;
```

This works, but it's a divergence from how formation/dive phases handle it — those go through `bossBurstFire`, which already knows it's a Boss. If a third tier ever needed custom speed in circle phase, this branch would be the only place to remember to update. Not a blocker, but worth noting.

---

### Potential Issues

- **Burst count comment is slightly misleading.** The `burstShotsLeft` value returned is the count *after* the shot being fired now. The comment "pick 3–5 total, return remaining" could be read as returning the full count. Consider: *"start new burst: pick 3–5 total shots; return remaining count after this shot."*

- **No test for bullet speed.** `BOSS_BULLET_VY = 0.46` is now a gameplay-critical constant, but nothing in the test suite asserts that boss bullets actually carry that velocity. A regression (accidentally removing the `BOSS_BULLET_VY` argument) would be invisible to CI.

- **`BURST_PAUSE_JITTER` interaction.** Post-burst pause is now `2000 + rng() * 1000`, so 2–3 s, not a flat 2 s. Not a bug — just worth being explicit about in the description for future balance tuning reference.

---

### Test Coverage

The `burstShotsLeft <= 4` update is correct and necessary. Two gaps:

1. **Missing: bullet speed assertion.** A test asserting a Boss bullet has `vy === BOSS_BULLET_VY` (and Elite has `vy === BULLET_E_VY`) would protect this going forward.
2. **Missing: circle phase coverage.** The new `enemy.tier === "Boss"` branch in `tickCircling` has no dedicated test.

---

### Verdict

Solid, well-scoped change with correct logic. Two suggestions before merging:

1. Add a test asserting boss bullet `vy === BOSS_BULLET_VY` in at least the burst path; ideally in the circle path too.
2. Clarify the inline burst count comment to say "return remaining after this shot" to avoid the off-by-one confusion.

Issue #1463, opened 2026-05-10
Star Swarm: fix Challenging Stage — lower HP, slower swarm, conditional scoring, no power-up


wcmchenry3-stack:
## Problem

The Challenging Stage as implemented fails on every design axis:

1. **Swarm exits too fast** — 40 enemies stagger in at 80 ms apart, each on a 3200 ms arc. The entire stage is over in ~6.3 seconds. Multi-HP enemies have a 2–3 second window on screen; there's no realistic time to engage them.
2. **HP too high for a bonus stage** — Boss (4 HP), Elite (2 HP) in a shooting gallery that's meant to be a skill showcase. The original Galaga uses 1-hit kills across the board (Boss = 2 hits as the premium target).
3. **Passive play beats active play** — the wave-clear bonus (`wave × 500`) is unconditional. At wave 3 the player scores 1,500 pts by standing still. Landing shots on a fleeting 4-HP Boss barely improves that — the math doesn't reward effort.
4. **Power-up is irrelevant — and should be removed** — The Lightning power-up spawns at the top and falls at 0.08 px/ms; it physically arrives ~530 ms after the last enemy exits. But more importantly: any power-up (Lightning, Shield, Bomb) trivialises the perfect-clear challenge. The whole point of the stage is that the player earns it with skill — a bomb or lightning burst undercuts that.

## Desired Behaviour

The Challenging Stage should be a genuine shooting-gallery skill test:
- Fast enough that it requires active engagement
- Slow enough that a skilled player can plausibly clear all enemies
- No free tools — the player wins or loses the perfect bonus on their own

### Enemy HP
- [ ] Grunt HP in Challenging Stage: **1** (unchanged)
- [ ] Elite HP in Challenging Stage: **1** (down from 2)
- [ ] Boss HP in Challenging Stage: **2** (down from 4)

### Pacing
- [ ] Path duration extended so enemies spend meaningfully more time on screen — total stage duration ~20–25 seconds
- [ ] Stagger interval tuned so the swarm spreads out visually without gaps feeling empty

### Scoring
- [ ] Wave-clear bonus made conditional on hit count: `Math.round((challengingHits / CHALLENGING_ENEMY_COUNT) * wave * 500 * difficultyMult)` — zero kills = zero wave-clear bonus
- [ ] Per-kill and per-hit bonuses unchanged
- [ ] Perfect-clear bonus (10,000 × difficulty) unchanged — still the stretch goal

### Power-up
[...]

PR #1465, opened 2026-05-10
feat(starswarm): rework Challenging Stage — lower HP, slower swarm, conditional scoring, no power-up (#1463)

wcmchenry3-stack:
Enemy HP reduced to Grunt 1 / Elite 1 / Boss 2 so the stage is a real
shooting gallery. Path duration extended from 3200 ms to 5000 ms; stagger
from 80 ms to 400 ms — total stage duration ~20.6 s (was 6.3 s). Wave-clear
bonus is now proportional to hit fraction (zero kills = zero bonus) so passive
play no longer scores nearly as much as active play. Power-up removed — any
bomb or lightning trivialises the perfect-clear target.

https://claude.ai/code/session_01V81eUam5bZoXudJSv2c8Zw

### DIFF
--- a/frontend/src/game/starswarm/__tests__/engine.test.ts
+++ b/frontend/src/game/starswarm/__tests__/engine.test.ts
@@ -738,10 +738,10 @@ describe("ChallengingStage off-screen cleanup", () => {
     let s = initStarSwarm(CANVAS_W, CANVAS_H, 3);
     expect(s.phase).toBe("ChallengingStage");
 
-    // With 40 enemies, last enemy (idx 39) has delay = 39*80/3200 = 0.975.
-    // It exits the canvas after (1 + 0.975) * 3200 ≈ 6320 ms.
+    // With 40 enemies, last enemy (idx 39) has delay = 39*400/5000 = 3.12 (×pathDuration).
+    // It exits the canvas after 39*400 + 5000 = 20600 ms.
     // Advance past that with no firing so enemies scroll off instead of being shot.
-    s = advanceMs(s, 7000, NO_INPUT);
+    s = advanceMs(s, 22000, NO_INPUT);
     expect(s.phase).toBe("WaveClear");
   });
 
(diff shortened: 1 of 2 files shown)


wcmchenry3-stack (commented):
## Code Review

### Code Quality

**`CHALLENGING_TIER_HP: Record<EnemyTier, number>` is the right type.**
Using `Record<EnemyTier, number>` means TypeScript will error if a new tier is ever added without updating this lookup — better than a partial object or a switch.

**Delay formula is correct.**
`(idx * CHALLENGE_STAGGER_MS) / CHALLENGE_PATH_DURATION` gives the normalized path offset, which when multiplied back through the tick loop yields `39 * 400 + 5000 = 20,600 ms` for the last enemy. Matches the PR description and updated test values exactly.

**Proportional bonus math is clean.**
```ts
const hitFraction = state.challengingHits / CHALLENGING_ENEMY_COUNT;
const waveClearBonus = Math.round(hitFraction * state.wave * WAVE_CLEAR_BONUS_BASE * sm);
```
Zero kills → zero bonus, full kills → same as the old flat bonus. `Math.round` is applied to the whole expression so partial clears get a sensible rounded value.

**Power-up removal is correct.**
`const powerUps: PowerUp[] = []` and deleting the `challengingPowerUp` block eliminates the `Math.random()` non-determinism that was only cosmetic anyway. The rest of the drop logic (`dropJitterTarget`, normal wave power-ups) is untouched. ✓

---

### Potential Issues

- **`hitFraction > 1.0` is unguarded.** If `challengingHits` somehow exceeds `CHALLENGING_ENEMY_COUNT` (e.g., a Boss is credited more than once due to a hit-counting bug), the bonus silently exceeds the intended cap. A one-line guard — `Math.min(1, state.challengingHits / CHALLENGING_ENEMY_COUNT)` — would make the formula defensive at negligible cost.

- **Naming inconsistency.** New constants use the `CHALLENGE_` prefix (`CHALLENGE_PATH_DURATION`, `CHALLENGE_STAGGER_MS`) while existing ones use `CHALLENGING_` (`CHALLENGING_ENEMY_COUNT`, `CHALLENGING_TIER_HP`, `challengingHits`, `challengingPerfect`). Minor, but worth aligning for grepping.

- **`advanceMs(s, 22000)` in tests.** 22 s of simulated time at 16 ms ticks = ~1,375 ticks per test call, and two tests do this. Probably fine in CI, but if the suite starts timing out this is the first place to look.

---

### Test Coverage

**Correctly updated:**
- Off-screen cleanup sentinel bumped from 7,000 ms → 22,000 ms.
- Power-up test inverted cleanly (`length === 1` → `length === 0`).
- `noPerfect.score === 0` is a new, explicit assertion that directly validates the zero-kills behavior — good addition.
- Perfect score delta now includes `waveClearBonus` in the expected value (`3 * 500 + 40 * 50 + 10_000 = 13,500`), correctly reflecting that `noPerfect.score` is now 0 rather than 1,500.

**Missing:**
- No test for an intermediate hit fraction (e.g., 20 kills → 50% wave-clear bonus). The proportional formula is only verified at the extremes (0 and 40 hits). A mid-point case would directly exercise the multiplication and rounding.

---

### Verdict

Solid, well-reasoned rework. Three things before merging:

1. Add `Math.min(1, ...)` guard to `hitFraction` to cap the bonus defensively.
2. Align constant prefix to `CHALLENGING_` for consistency with the existing codebase vocabulary.
3. Add a partial-hit test case (e.g., 20 kills) to cover the proportional scoring formula at an intermediate value.

Issue #1502, opened 2026-05-10
feat(hearts-sim): add per-hand and per-trick diagnostic logging to the Hearts simulator


wcmchenry3-stack:
## Context

The existing simulator (`scripts/simulate-hearts.ts`) runs 3,000-game batches and emits aggregate stats: win rate, mean score, moon shots, Q♠-on-human frequency. This is useful for broad balance checks but gives no visibility into individual AI decisions.

Investigating the bugs in #1500 and #1501 — and catching similar issues in the future — requires being able to inspect what actually happened on a per-trick level without manually playing through games.

## Requested feature

Extend the simulator to optionally emit a structured **game log** with:

**Per hand:**
- Initial 13-card deal for each player
- 3 cards each player passed (and direction)
- 3 cards each player received

**Per trick:**
- Which player led
- Card played by each player, in play order
- Trick winner

**Output format:** newline-delimited JSON (one object per game, or one object per trick — whichever is easier to `grep`/`jq`). Should not pollute the existing aggregate output.

**Suggested interface:**

```
npx tsx scripts/simulate-hearts.ts --log-games 10
```

Runs 10 fully logged games (instead of the 3,000-game aggregate batch) and writes JSON to stdout or a file.

## Why this is useful

With structured logs you can answer questions like:

- "Find all tricks where an AI held K♠ and led it while Q♠ was not yet played."
- "How often does hard AI attempt a moon shot and succeed vs. fail?"
- "How many times did an AI pass Q♠ to the human vs. keeping it?"

Currently these require a separate debugging session or adding temporary `console.log` to the engine.

## Affected files

- `scripts/simulate-hearts.ts` — extend `GameResult` / add `GameLog` type, capture state snapshots in the game loop
- No engine or AI changes needed; all required data is already present in `HeartsState` after each `playCard` / `commitPass` call

Issue #1501, opened 2026-05-10
bug(hearts-ai): Medium AI leads K♠ when Q♠ is still in play


wcmchenry3-stack:
## Observed behavior

A medium-difficulty AI player (George) led K♠ when other safe cards were available and Q♠ had not yet been played. This is a risky lead — any opponent holding Q♠ can dump it on the K♠ trick, giving George 13 points.

## Root cause

`chooseLead` in `frontend/src/game/hearts/ai.ts` (line 500) filters hearts and Q♠ itself from the safe pool, but does **not** filter K♠ or A♠:

```typescript
function chooseLead(valid: Card[]): Card {
  // Avoid leading hearts or Q♠ unless forced
  const safe = valid.filter((c) => c.suit !== "hearts" && !isQueenOfSpades(c));
  // K♠ and A♠ remain in `safe` — no guard
  ...
}
```

Hard difficulty avoids this via `chooseLeadHard` (line 520), which excludes K♠/A♠ from the safe pool until Q♠ has been seen in `wonCards` or `currentTrick`:

```typescript
if (c.suit === "spades" && (c.rank === 13 || c.rank === 1) && !qSpadeGone) return false;
```

Medium has no equivalent guard.

## Expected behavior

Medium AI should avoid leading K♠ or A♠ when Q♠ is still live. Medium does not need full card counting — a simple derivation suffices: if Q♠ is not in this player's hand **and** not present in `state.wonCards` or `state.currentTrick`, treat K♠/A♠ as risky and exclude them from the safe lead pool (fall back to them only if no other option exists).

## Affected files

- `frontend/src/game/hearts/ai.ts` — `chooseLead` (line 500); `chooseLeadHard` (line 520) for reference
- `frontend/src/game/hearts/__tests__/ai.test.ts` — add regression tests

Issue #1500, opened 2026-05-10
bug(hearts-ai): AI plays lowest following card when it can't win the trick (should play highest losing card)


wcmchenry3-stack:
## Observed behavior

When following a led suit and the AI cannot win the trick, it plays its **lowest** card in suit. This misses the opportunity to exhaust dangerous high cards — most notably dumping Q♠ when K♠ is the current trick winner.

**Concrete example:** George leads K♠. Kramer holds Q♠ (rank 12) and several lower spades. Since K♠ (rank 13) beats Q♠, playing Q♠ here would safely lose the trick and rid Kramer of the 13-point liability. Instead the AI plays its lowest spade and keeps Q♠ for a worse moment.

## Root cause

`chooseFollow` in `frontend/src/game/hearts/ai.ts` (line 543) has two separate paths:

```typescript
if (pts > 0) {
  // Trick has points — try to lose
  if (losing.length > 0) {
    return highest(losing) ?? valid[0]!;  // ✅ highest card that still loses
  }
  return lowest(inSuit) ?? valid[0]!;
}

// No points, not last — play lowest to stay safe
return lowest(inSuit) ?? valid[0]!;        // ❌ should also be highest(losing)
```

The `pts === 0, not last` branch uses `lowest` instead of `highest(losing)`. When you can't win the trick, exhausting your highest non-winning cards is always correct — whether the trick currently carries points or not.

## Expected behavior

When following suit and unable to win the trick, play the **highest card that still loses** (`highest(losing)`). This unifies the two branches and makes Q♠-on-K♠ dumping a natural consequence.

## Exception: moon-attempt mode

Hard difficulty (`selectCardToPlayHard`, line 374) enters moon-attempt mode when the player holds 8+ hearts + Q♠ with no points yet taken. In that mode the player needs Q♠ to complete the shot — the `chooseFollow` fix must be gated so it does not apply when `isMoonAttempt` is true. Medium and Easy have no moon-attempt mode so no exception is needed there.

## Affected files

- `frontend/src/game/hearts/ai.ts` — `chooseFollow` (line 543), `selectCardToPlayHard` (line 374)
- `frontend/src/game/hearts/__tests__/ai.test.ts` — add regression tests

PR #1506, opened 2026-05-11
fix(hearts): prevent AI loop freeze when human taps card during trick animation (#1505)

### LINKED ISSUE #1505: Hearts: game freezes in limbo when human taps card during trick animation
## Summary

After an AI-completed trick where the human wins, the game can freeze: the human's card appears on the table but the three CPU players never follow, leaving the trick permanently incomplete.

## Reproduction

1. Play a hand of Hearts until the AI players complete a trick that the human wins (the 4-card "trick taken" animation starts playing).
2. **Before the animation finishes**, tap one of the human player's remaining cards (which are already highlighted as valid plays).
3. The human's card appears on the table, but the three AI players never play their cards. The game is stuck.

This is easiest to trigger late in a hand when the human leads the final trick — there is only one card left in hand, so the impulse to tap it quickly is natural.

## Root Cause

`HeartsScreen.tsx` has an async AI turn loop (`runAiTurns`) guarded by `loopActiveRef.current`. When an AI-completed trick resolves to a human win, the loop:

1. calls `setLastTrick({...})` to start the 4-card animation, and
2. immediately `await`s a `Promise` stored in `trickAnimResolverRef.current`, keeping `loopActiveRef.current === true` until `handleTrickAnimationComplete` fires.

The invariant is that `handleTrickAnimationComplete` will resolve the promise and clear `lastTrick`. However, `handleCardPress` did **not** guard against `lastTrick !== null`. Because `gameState.currentPlayerIndex` was already set to `HUMAN` (the trick winner), the human's valid cards were shown and interactive. Tapping one called `setLastTrick(null)` — clearing the animation — **without** ever resolving the pending promise.

Result:

- `trickAnimResolverRef.current` is now an orphaned, never-resolving promise.
- `loopActiveRef.current` stays `true` forever.
- Every subsequent call to `runAiTurns` hits the early-return guard and does nothing.
- The game is visually stuck: one card on the table, zero CPU plays.

```
// HeartsScreen.tsx — handleCardPress (BEFORE fix)
function handleCardPress(card: Card) {
  if (!gameState || gameState.currentPlayerIndex !== HUMAN || gameState.phase !== "playing")
    return;
  // ← No guard for lastTrick !== null. Human CAN tap while animation runs.
  ...
  setLastTrick(null);   // ← clears animation WITHOUT resolving the pending promise
  setGameState(newState);
}
```

## Fix

`handleCardPress` now returns early when `lastTrick !== null`. The human must wait for the trick-taken animation to complete naturally before playing. Once `handleTrickAnimationComplete` fires it resolves the promise, the AI loop exits cleanly (`loopActiveRef.current = false`), and the next `runAiTurns` invocation starts normally.

```diff
 function handleCardPress(card: Card) {
   if (!gameState || gameState.currentPlayerIndex !== HUMAN || gameState.phase !== "playing")
     return;
+  // Don't accept input while the completed-trick animation is still playing.
+  // If the human taps during this window and clears lastTrick, the AI loop's
+  // pending animation-resolver promise would never be fulfilled, leaving
+  // loopActiveRef.current === true and freezing all subsequent AI turns.
+  if (lastTrick !== null) return;
   ...
 }
```

**File:** `frontend/src/screens/HeartsScreen.tsx`

## Tests Added

A new `describe` block in `frontend/src/screens/__tests__/HeartsScreen.test.tsx`:

> **"HeartsScreen — AI loop frozen regression (race condition)"**

The test loads a game state where trick 13 is already in progress (human has played the leading ♥5, all three AIs have one diamond each left to follow), advances fake timers by 2 000 ms (covering the three 400 ms AI delays), and asserts that the **"Hand Complete"** modal appears. Before the fix, if this state was reached after a mid-animation card tap, `loopActiveRef` would be stuck and the test would time out waiting for the modal.

## Impact

- Any hand where the human wins an AI-completed trick (common) and taps their next card before the ~1–2 s animation finishes.
- Affects all AI difficulty levels.
- The only recovery is to restart the game.

wcmchenry3-stack:
Closes #1505

handleCardPress now returns early when `lastTrick !== null` (trick-taken animation in progress). Previously the human could tap before the animation fired `handleTrickAnimationComplete`, which called `setLastTrick(null)` without resolving the AI loop's pending Promise — leaving `loopActiveRef.current` true forever and freezing all subsequent CPU turns.

Adds a regression test: loads trick-13-in-progress state (human led, 3 AIs to follow), advances fake timers, asserts the "Hand Complete" modal appears.

https://claude.ai/code/session_01D5HqXmBnb279WP7WW4FMTS

### DIFF
--- a/frontend/src/screens/HeartsScreen.tsx
+++ b/frontend/src/screens/HeartsScreen.tsx
@@ -301,6 +301,11 @@ export default function HeartsScreen() {
   function handleCardPress(card: Card) {
     if (!gameState || gameState.currentPlayerIndex !== HUMAN || gameState.phase !== "playing")
       return;
+    // Don't accept input while the completed-trick animation is still playing.
+    // If the human taps during this window and clears lastTrick, the AI loop's
+    // pending animation-resolver promise would never be fulfilled, leaving
+    // loopActiveRef.current === true and freezing all subsequent AI turns.
+    if (lastTrick !== null) return;
     ensureSyncStarted();
     if (isQueenOfSpades(card)) {
       humanJustPlayedQSRef.current = true;
--- a/frontend/src/screens/__tests__/HeartsScreen.test.tsx
+++ b/frontend/src/screens/__tests__/HeartsScreen.test.tsx
@@ -6,6 +6,7 @@ import { HeartsRoundsProvider } from "../../game/hearts/RoundsContext";
 import { createSeededRng, setRng } from "../../game/hearts/engine";
 import * as engine from "../../game/hearts/engine";
 import { loadGame } from "../../game/hearts/storage";
+import type { HeartsState } from "../../game/hearts/types";
 
 jest.mock("../../game/hearts/storage", () => ({
   loadGame: jest.fn().mockResolvedValue(null),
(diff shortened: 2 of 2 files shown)


wcmchenry3-stack:
Superseded by #1508, which expands scope to include #1500, #1501, #1502 and also fixes the Playwright E2E CI failure (MCR Docker image blocked).

Issue #1518, opened 2026-05-11
epic(hearts-analysis): local browser dashboard for Hearts game log analysis


wcmchenry3-stack:
## Overview

A standalone local dev tool that runs Hearts game simulations with configurable parameters, then presents strategic analysis in a browser dashboard. FastAPI backend, plain HTML/JS frontend with Chart.js — no build step, no deployment concerns.

Not part of the BC Arcade app. Lives in `hearts-analysis/` in this repo.

## Run

```
cd hearts-analysis
pip install -r requirements.txt
uvicorn main:app --reload
# open http://localhost:8000
```

Use the dashboard's simulation controls to set game count and per-player difficulties, then click "Run Simulation."

### Simulator extension
- [ ] #1512 — Add `--count` / `--difficulties` flags to `simulate-hearts.ts`

### Backend
- [ ] #1513 — Scaffold `hearts-analysis/` (FastAPI + static HTML stub)
- [ ] #1514 — `POST /api/simulate` endpoint (invokes TS simulator, reloads data)
- [ ] #1515 — Implement all `analyzer.py` analysis functions
- [ ] #1516 — Wire all `GET /api/*` endpoints in `main.py`

### Frontend
- [ ] #1517 — Chart.js dashboard in `static/index.html` (all 8 sections + simulation controls)

## Analysis sections

| Section | Chart type | Key insight |
|---|---|---|
| Summary cards | Cards | Avg score, wins, pts/hand, Q♠ taken — color coded |
| Score by game | Line chart | Per-player game-by-game trajectory |
| Safe discard rate | Stacked bar | Hearts/Q♠ dumped vs wasted void opportunities |
| Q♠ dump timing | Line chart | Distribution of which trick Q♠ leaves hand |
| Moon eligibility | Stacked bar | Eligible hands vs actual damage taken |
| Pass direction | Grouped bar | Avg pts per player per direction |
| Missed Q♠ dumps | Table | Specific incidents where AI failed to dump Q♠ |
| Simulation controls | Form | Game count (1–500) + per-player difficulty selects |

## Sanity check (seeds 0–9, all medium)

When the tool is working, 10 games with seeds 0–9 and all-medium difficulties should reproduce:
- P3: 4 wins, avg 64.0, fewest Q♠ taken (20)
- P2: 2 wins, avg 77.7, most Q♠ taken (34)
- P0 safe discard rate: 66.4% (worst)
[...]

Issue #1517, opened 2026-05-11
feat(hearts-analysis): implement Chart.js analysis dashboard in static/index.html


wcmchenry3-stack:
## Context

The dashboard is a single HTML file with no build step. It fetches all API endpoints in parallel on load and renders the results using Chart.js from cdnjs.

Depends on: #1516 (endpoints wired), #1514 (simulate endpoint)

### 1. Simulation controls
- "Number of games" number input (default 10, min 1, max 500)
- Four difficulty selects (P0–P3), options: easy / medium / hard (default medium)
- "Run Simulation" button → `POST /api/simulate`, shows loading state, reloads all data on completion
- Error display if simulate fails

### 2. Summary cards
4 cards per player, color-coded (red = worst, green = best across players):
- Avg final score (lower = better)
- Games won
- Avg pts/hand
- Q♠ taken (total)

### 3. Score by game (line chart)
- X-axis: game index 1–N
- Y-axis: final score (lower = better)
- One line per player, different dash patterns for accessibility
- Chart.js from cdnjs (no npm)

### 4. Safe discard rate (stacked bar)
- Per player: hearts dumped / Q♠ dumped / wasted stacks
- Show waste rate % as tooltip annotation

### 5. Q♠ dump timing (line chart)
- X-axis: trick 1–13, Y-axis: frequency
- One line per player — shows distribution shape

### 6. Moon eligibility (stacked bar)
- Outcome buckets per player: 0pts / 1–5 / 6–12 / 13–19 / 20–26
- Section header notes if zero moon shots were made

### 7. Pass direction (grouped bar)
- Avg pts per player per direction (left/right/across/none)
- Highlights standout cells

### 8. Missed Q♠ dump table
- Columns: game, hand, trick, player, covering card, played instead, pts taken that hand
- Client-side sortable by pts taken

## Technical constraints

- Single `.html` file, no build step, no npm
- Chart.js loaded from cdnjs
- Fetches all endpoints via `Promise.all` on page load
- Re-fetches all after simulate completes

## Acceptance criteria

- All 8 sections render without errors on the 10-game dataset
- Sanity check values (from #1515) are visually apparent
- "Run Simulation" button generates new data and the dashboard updates
- Page is usable at standard laptop screen width

## Affected files

- `hearts-analysis/static/index.html`

wcmchenry3-stack:
Completed in PR #1519 (merged).

Issue #1513, opened 2026-05-11
feat(hearts-analysis): scaffold hearts-analysis/ local dev tool


wcmchenry3-stack:
## Context

We need a standalone local dev tool (separate from the main BC Arcade app) for analyzing Hearts game logs. This issue scaffolds the directory structure and wires up the basics.

## Directory layout

```
hearts-analysis/
├── main.py             # FastAPI app + all API endpoints
├── analyzer.py         # pure analysis functions (no side effects)
├── requirements.txt    # fastapi, uvicorn
├── data/               # game log JSON files (git-ignored or empty)
│   └── .gitkeep
└── static/
    └── index.html      # dashboard (stub — full implementation in separate issue)
```

## Requirements

**`requirements.txt`**
```
fastapi
uvicorn[standard]
```

**`main.py`** — stub only:
- `GAMES` global loaded at startup from `data/`
- Placeholder endpoints returning `{"ok": true}` for:
  `GET /api/summary`, `GET /api/discard-rate`, `GET /api/qs-timing`,
  `GET /api/moon`, `GET /api/pass-direction`, `GET /api/missed-dumps`
- `POST /api/reload` — re-reads `data/` without restarting
- `POST /api/simulate` — invokes `npx tsx ../scripts/simulate-hearts.ts` with body params `{ count, difficulties }`, writes output to `data/games.json`, then reloads (see issue #1514)
- `GET /` — serves `static/index.html`

**`static/index.html`** — minimal stub with "Dashboard coming soon" placeholder

## Run instructions (in issue body, not code)

```
cd hearts-analysis
pip install -r requirements.txt
uvicorn main:app --reload
# open http://localhost:8000
```

## Acceptance criteria

- `uvicorn main:app --reload` starts without errors
- `GET /` returns the HTML stub
- All `/api/*` stubs return 200
- `data/` directory exists with `.gitkeep`

## Affected files

New: `hearts-analysis/` (entire directory)

wcmchenry3-stack:
Completed in PR #1519 (merged).

Issue #1512, opened 2026-05-11
feat(hearts-sim): add --difficulties and --count flags to simulate-hearts.ts


wcmchenry3-stack:
## Context

The simulator currently hardcodes all four players as `medium` when `--log-games N` is used. The upcoming Hearts Analysis Tool needs to run simulations with configurable player difficulties and game counts from a browser UI.

## Requested changes

**`scripts/simulate-hearts.ts`**

Add two new CLI flags to the `--log-games` path:

| Flag | Type | Default | Description |
|---|---|---|---|
| `--count N` | int | required | Number of games to simulate and log |
| `--difficulties p0,p1,p2,p3` | comma-separated | `medium,medium,medium,medium` | Per-player difficulty (`easy`/`medium`/`hard`) |

Rename `--log-games N` → `--count N` for clarity, keeping `--log-games` as a deprecated alias so existing scripts don't break.

**Example usage:**
```
npx tsx scripts/simulate-hearts.ts --count 10 --difficulties medium,medium,hard,easy
```

Output: NDJSON to stdout, one game object per line (same format as current `--log-games`).

## Acceptance criteria

- `--count 5 --difficulties easy,hard,medium,medium` runs 5 logged games with the specified per-player difficulties
- `difficulties` field in each game JSON reflects the actual values passed (not hardcoded `medium`)
- `--log-games N` still works (backward compat)
- Invalid difficulty value exits with a clear error message
- No changes to the existing aggregate-stats batch path

## Affected files

- `scripts/simulate-hearts.ts`

wcmchenry3-stack:
Completed in PR #1519 (merged).

PR #1519, opened 2026-05-11
feat(hearts-analysis): local browser dashboard for Hearts game log analysis

wcmchenry3-stack:
## Summary

Implements the full Hearts Analysis Tool epic (#1518) — a local dev tool that runs Hearts simulations and displays strategic analysis in a browser dashboard.

- **#1512** — `simulate-hearts.ts`: new `--count N` and `--difficulties p0,p1,p2,p3` CLI flags; `--log-games` kept as deprecated alias
- **#1513/#1514/#1515/#1516** — `hearts-analysis/` FastAPI backend: `load_games`, `summary_stats`, `safe_discard_rate`, `qs_dump_timing`, `moon_eligibility`, `pass_direction_breakdown`, `missed_qs_dumps`; `POST /api/simulate` writes simulator NDJSON directly to file (avoids pipe-buffer truncation on large outputs)
- **#1517** — `static/index.html`: Chart.js dashboard with simulation controls (game count + per-player difficulty dropdowns), summary metric cards, 5 charts, and a sortable missed-dump table
- **chore** — exempts `static/` dirs from the design-tokens policy gate (dev tool dashboards use inline styles by necessity)

## Test plan

- [ ] `cd hearts-analysis && pip install -r requirements.txt && uvicorn main:app --reload`
- [ ] Open `http://localhost:8000`, set game count + difficulties, click "Run Simulation"
- [ ] Verify all 8 dashboard sections render with 10 all-medium games
- [ ] Sanity check: P3 should have the most wins and lowest avg score; P2 most Q♠ taken
- [ ] `npx tsx scripts/simulate-hearts.ts --count 5 --difficulties easy,hard,medium,easy` → 5 NDJSON lines with correct difficulties field
- [ ] `--log-games 3` still works (backward compat)

### DIFF
--- a/.claude/policies/policy-patterns.json
+++ b/.claude/policies/policy-patterns.json
@@ -17,6 +17,6 @@
   },
   "design-tokens": {
     "detect": ":\\s*#[0-9a-fA-F]{3,8}|rgba?\\s*\\(|hsla?\\s*\\(|font-size:\\s*[0-9]+px|font-family:\\s*[\"']|tabindex=[\"'][1-9]",
-    "skip": "node_modules/|(^|/)dist/|(^|/)build/|(^|/)\\.cache/|(^|/)coverage/|(^|/)public/|(^|/)tokens/|\\.min\\.(css|js)|\\.stories\\.[jt]sx?|\\.test\\.[jt]sx?|\\.spec\\.[jt]sx?|__snapshots__|\\.svg$|\\.mdx?$|\\.tokens\\.json$|tailwind\\.config\\.|theme\\.[^./]+\\.[jt]sx?$|(^|/)theme/"
+    "skip": "node_modules/|(^|/)dist/|(^|/)build/|(^|/)\\.cache/|(^|/)coverage/|(^|/)public/|(^|/)static/|(^|/)tokens/|\\.min\\.(css|js)|\\.stories\\.[jt]sx?|\\.test\\.[jt]sx?|\\.spec\\.[jt]sx?|__snapshots__|\\.svg$|\\.mdx?$|\\.tokens\\.json$|tailwind\\.config\\.|theme\\.[^./]+\\.[jt]sx?$|(^|/)theme/"
   }
 }
--- a/.claude/policies/policy-patterns.json
+++ b/.claude/policies/policy-patterns.json
@@ -17,6 +17,6 @@
   },
   "design-tokens": {
     "detect": ":\\s*#[0-9a-fA-F]{3,8}|rgba?\\s*\\(|hsla?\\s*\\(|font-size:\\s*[0-9]+px|font-family:\\s*[\"']|tabindex=[\"'][1-9]",
-    "skip": "node_modules/|(^|/)dist/|(^|/)build/|(^|/)\\.cache/|(^|/)coverage/|(^|/)public/|(^|/)static/|(^|/)tokens/|\\.min\\.(css|js)|\\.stories\\.[jt]sx?|\\.test\\.[jt]sx?|\\.spec\\.[jt]sx?|__snapshots__|\\.svg$|\\.mdx?$|\\.tokens\\.json$|tailwind\\.config\\.|theme\\.[^./]+\\.[jt]sx?$|(^|/)theme/"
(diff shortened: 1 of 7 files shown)


wcmchenry3-stack:
## PR #1519 — Hearts Analysis Dashboard

**Overview:** Implements a complete local dev tool for analyzing Hearts AI games — a FastAPI backend with 7 analysis functions and a Chart.js dashboard. One commit covers all 6 issues in the epic. The core architecture is solid and the pipe-buffer fix for large subprocess output is a good catch.

---

### Bugs

**1. Table sort breaks on object-valued columns** (`static/index.html:1021`)

\`\`\`js
const sorted = [...missedData].sort((a, b) => sortDir * (b[sortCol] - a[sortCol]));
\`\`\`

When a user clicks "Covering Card" or "Played Instead", \`b[sortCol]\` is an object \`{suit, rank}\`. Subtracting two objects produces \`NaN\`, silently breaking the sort. All rows stay in their current order with no feedback. Either disable \`data-col\` on those two headers or sort by \`rank\` when the column value is an object.

**2. \`loadAll\` doesn't include \`/api/game-scores\`** (`static/index.html:800–817`)

\`renderScoreChart\` makes its own separate fetch inside \`Promise.all(...).then(...)\`. This is an extra round-trip and, more importantly, it's fire-and-forget with a \`.catch(() => {})\`. If the second fetch fails silently the score chart stays empty with no error visible. The endpoint should be included in the \`Promise.all\` in \`loadAll\`.

---

### Dead Code

**3. \`held_qs\` set in \`qs_dump_timing\` is computed but never used** (`analyzer.py:253–259`)

\`\`\`python
held_qs = {
    p for p in range(4)
    if any(_is_qs(c) for c in received[p])
    or any(_is_qs(c) for c in hand.get("initialDeal", ...)[p])
    if not hand.get("passed", ...)[p]   # ← second `if` is a filter, unusual
    or not any(_is_qs(c) for c in hand.get("passed", ...)[p])
}
\`\`\`

The next block recomputes \`qs_holders\` from scratch via the correct loop. \`held_qs\` is entirely unreferenced. Delete it.

**4. \`penalty\` variable in \`safe_discard_rate\` is assigned but never read** (`analyzer.py:211–214`)

\`\`\`python
penalty = _card_points(card) > 0
if trick_number == 1 and (_is_heart(card) or _is_qs(card)):
    penalty = False   # written, never read again
\`\`\`

The branching below uses \`trick_number != 1\` directly instead of \`penalty\`. Either use \`penalty\` to drive the if/elif or delete both lines.

**5. \`CARD_RANK_MAP\` is defined but never used** (`analyzer.py:36`)

\`_rank()\` has its own inline \`14 if r == 1 else r\` logic. The constant is dead. Remove it.

**6. Stale comment in \`renderScoreChart\`** (`static/index.html:864–869`)

The five-line comment explains why a \`/api/scores\` endpoint was needed — but \`/api/game-scores\` now exists and is used. The comment is obsolete and confusing.

---

### Minor Concerns

**7. \`@app.on_event("startup")\` is deprecated** (`main.py:499`)

FastAPI ≥ 0.93 prefers the \`lifespan\` context manager. Low risk for a dev tool but worth flagging for when dependencies get upgraded:

\`\`\`python
from contextlib import asynccontextmanager

@asynccontextmanager
async def lifespan(app: FastAPI):
    _reload_games()
    yield

app = FastAPI(lifespan=lifespan)
\`\`\`

**8. \`POST /api/simulate\` blocks the event loop** (`main.py:589`)

\`subprocess.run\` is synchronous. For 500 games this can take tens of seconds during which the server is unresponsive. Acceptable for a single-user dev tool, but if it ever runs more than one request concurrently, the second request stalls. \`asyncio.to_thread(subprocess.run, ...)\` would fix it without adding complexity.

**9. No subprocess timeout** (`main.py:589–595`)

A runaway \`npx\` invocation (e.g. due to an engine infinite loop) would hang the server indefinitely. Adding \`timeout=120\` to \`subprocess.run\` is a one-liner safety net.

**10. Policy exemption is broad** (`.claude/policies/policy-patterns.json`)

\`(^|/)static/\` skips design-token checks for *any* \`static/\` directory in the repo. This is correct for the dev tool but would silently exempt a \`frontend/static/\` if one ever appeared. Scoping it to \`hearts-analysis/static/\` would be safer:

\`\`\`
(^|/)hearts-analysis/static/
\`\`\`

---

### What's Good

- **Pipe buffer fix** — writing subprocess stdout directly to a file rather than capturing it in memory is the right call; the 64 KB truncation would have been a hard-to-diagnose bug.
- **\`missed_qs_dumps\` check ordering** — checking before updating \`covering_card\` correctly avoids the self-referential bug where a player's own K♠ play was being flagged as its own covering card.
- **Effective rank for Ace** — using \`_rank(card)\` (which maps rank 1 → 14) in the covering card comparison is correct; the initial version using \`card["rank"]\` would have silently skipped all A♠ covering cases.
- **Pydantic validation** on \`SimulateRequest\` — the length check and \`Literal\` type prevent injection of arbitrary strings into the subprocess command.
- **Clean separation** — \`analyzer.py\` is pure functions with no I/O; \`main.py\` owns all side effects. Easy to unit test.

PR #1523, opened 2026-05-11
fix(hearts): exclude Ace of Clubs from low-club pass guard in AI

### LINKED ISSUE #1522: Hearts: Ace of Clubs never passed by AI (passSafeFilter treats rank 1 as low club)
## Bug

In Hearts, the Ace of Clubs is effectively absent from the first half of every hand. It never moves in the passing phase and only surfaces in the final clubs tricks.

## Root Cause

Two compounding issues in `frontend/src/game/hearts/ai.ts`:

### 1. `passSafeFilter` incorrectly blocks the Ace from being passed

```typescript
// ai.ts:86 — before fix
if (c.suit === "clubs" && c.rank < 6) return false;
```

The Ace is stored as `rank === 1` (ranks run 1–13). Because `1 < 6` is true, the Ace of Clubs is caught by the same guard meant to protect low clubs (2♣–5♣). No AI player can ever include it in their pass selection at any difficulty.

### 2. AI play strategy defers the Ace until the last possible moment

When following a clubs lead, the AI calls `lowest(inSuit)` to pick its card. `lowest()` uses `aceHigh()` which maps rank 1 → 14, making the Ace the *highest* club. So it's always deferred until it's the only club remaining in the player's hand.

## Impact

- The Ace of Clubs is never included in any AI passing exchange.
- During trick-taking, the Ace is held back until the final clubs trick, making it invisible for most of the hand.

## Fix

Change `ai.ts:86` so the Ace (rank 1) is excluded from the low-club guard:

```typescript
// after fix
if (c.suit === "clubs" && c.rank > 1 && c.rank < 6) return false;
```

This matches the documented intent ("clubs below 6" = 2♣–5♣ in standard card notation) without accidentally catching the Ace.

wcmchenry3-stack:
Closes #1522

## Summary

- `passSafeFilter` in `ai.ts` used `c.rank < 6` to block AI from passing low clubs (2♣–5♣), but the Ace is stored as `rank === 1`, so `1 < 6` caught it too — the AI could never pass the Ace of Clubs
- Fix: change to `c.rank > 1 && c.rank < 6` so only ranks 2–5 are blocked, matching the stated intent "clubs below 6"

## Test plan

- [ ] Play a Hearts game and observe the Ace of Clubs appearing in the passing exchange in some hands
- [ ] Verify AI still never passes 2♣, 3♣, 4♣, or 5♣
- [ ] Verify no regression in existing Hearts engine and AI test suites

https://claude.ai/code/session_017MKw8jpucgf7NykzsNKuVT

---
_Generated by [Claude Code](https://claude.ai/code/session_017MKw8jpucgf7NykzsNKuVT)_

### DIFF
--- a/frontend/src/game/hearts/ai.ts
+++ b/frontend/src/game/hearts/ai.ts
@@ -83,7 +83,7 @@ function passSafeFilter(selected: Card[]): (c: Card) => boolean {
   return (c: Card) => {
     if (selected.some((s) => s.suit === c.suit && s.rank === c.rank)) return false;
     if (c.suit === "clubs" && c.rank === 2) return false;
-    if (c.suit === "clubs" && c.rank < 6) return false;
+    if (c.suit === "clubs" && c.rank > 1 && c.rank < 6) return false;
     return true;
   };
 }


wcmchenry3-stack:
The fix is scoped to clubs because `passSafeFilter` only has club-specific logic — it's the only suit that has a "never pass low cards" guard. The other suits don't have this problem:

- **Hearts**: The Ace is explicitly identified with `c.rank === 1` and is *prioritized* for passing (it's in the `dangerHearts` filter at `ai.ts:125`)
- **Spades**: The Ace is also explicitly checked with `c.rank === 1` in the lead-avoidance logic in `chooseLead`
- **Diamonds**: No special passing restrictions at all

The clubs filter was the only place using a numeric range comparison (`c.rank < 6`) rather than an explicit rank check, which is what caused the Ace to be silently swept up. Other suits never had that pattern.

So yes, the one-line change is the complete fix — it's not that other suits were missed, it's that the bug was structurally unique to how the clubs guard was written.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1526, opened 2026-05-12
fix(hearts-ai): dump Q♠ when last to play and K♠/A♠ covering (#1525)

### LINKED ISSUE #1525: fix(hearts-ai): Q♠ not dumped when last to play with K♠/A♠ covering
## Summary

In `chooseFollow` (`frontend/src/game/hearts/ai.ts`), the `pts === 0 && isLastToPlay` branch always excludes Q♠ from `safeInSuit` via `cardPoints(c) === 0`. The intent is to avoid dumping Q♠ when you'd win the trick yourself — but when K♠ or A♠ is already the current winner (`winningRank > 12`), Q♠ is a **losing card** that would safely land on the high-spade holder. The branch never checks this, so the AI plays a low spade instead of seizing the dump.

## Impact

Identified via the Hearts analysis simulator (100-game run). The missed-Q♠-dumps table shows many high-damage incidents caused specifically by this path:

- G14 H3 T3 — P2 holds Q♠, K♠ covering, played J♠ → **25 pts taken**
- G78 H5 T8 — P1 holds Q♠, K♠ covering, played 10♠ → **25 pts taken**
- G30 H4 T6 — P0 holds Q♠, A♠ covering, played 7♠ → **24 pts taken**
- G2 H8 T7/T8 — P0, 12 pts each trick
- Multiple 8–21 pt incidents across the 100-game corpus

The non-last-to-play branches (`pts > 0` and the final "not last" block) already handle this correctly via `losing.find(isQueenOfSpades)`. Only the `isLastToPlay` branch was missing the check.

## Fix

Add a Q♠ escape hatch before the `safeInSuit` filter:

```typescript
if (pts === 0 && isLastToPlay) {
  // If Q♠ is a losing card (K♠ or A♠ covering), dump it — it won't come back to us
  const qSpade = inSuit.find(isQueenOfSpades);
  if (qSpade && aceHigh(qSpade.rank) < winningRank) return qSpade;
  // Safe trick, last to play — exhaust high cards, but never dump a point card onto ourselves
  const safeInSuit = inSuit.filter((c) => cardPoints(c) === 0);
  if (safeInSuit.length > 0) return highest(safeInSuit) ?? valid[0]!;
  return lowest(inSuit) ?? valid[0]!;
}
```

Two regression tests added covering the dump-when-safe and no-dump-when-would-win cases.

wcmchenry3-stack:
## Summary

- Fixes a bug in `chooseFollow` where the `pts === 0 && isLastToPlay` branch never dumped Q♠, even when K♠ or A♠ was already winning the trick and Q♠ was a safe losing card
- Adds two regression tests: one confirming the dump fires when covered, one confirming Q♠ is held when it would win the trick itself

## Root cause

`safeInSuit = inSuit.filter((c) => cardPoints(c) === 0)` excluded Q♠ unconditionally. The intent ("don't dump points onto ourselves") is correct when you'd win — but not when a higher spade is covering. The non-last-to-play branches already handled this via `losing.find(isQueenOfSpades)`; this PR brings `isLastToPlay` into parity with a one-line guard.

## Impact (from 100-game simulator run)

High-damage incidents this fix eliminates:
- G14 H3 T3 → 25 pts (P2, K♠ covering, played J♠)
- G78 H5 T8 → 25 pts (P1, K♠ covering, played 10♠)
- G30 H4 T6 → 24 pts (P0, A♠ covering, played 7♠)
- G2 H8 T7+T8 → 12 pts each (P0)
- Multiple 8–21 pt incidents across the corpus

## Test plan

- [x] `npx jest --testPathPattern="hearts"` — 216 tests pass (214 existing + 2 new)
- [ ] Re-run simulator (100 games, all medium) and verify missed-dump table shrinks, especially high-damage rows

Closes #1525

### DIFF
--- a/frontend/src/game/hearts/__tests__/ai.test.ts
+++ b/frontend/src/game/hearts/__tests__/ai.test.ts
@@ -927,4 +927,42 @@ describe("chooseLead — medium AI avoids risky spade leads (#1501)", () => {
     const pick = selectCardToPlay(hand, [], state, 0, "medium");
     expect(pick).toEqual(c("spades", 13));
   });
+
+  it("dumps Q♠ when last to play and K♠ is covering (safe dump opportunity)", () => {
+    // P3 is last to play. A♠ led the trick, K♠ and 9♠ already played (K♠ covers Q♠).
+    // P3 holds Q♠ and 7♠. Q♠ should be dumped — K♠ takes the trick, not Q♠.
+    const hand = [c("spades", 12), c("spades", 7)];
+    const trick: TrickCard[] = [
+      { card: c("spades", 1), playerIndex: 0 },  // A♠ led — winningRank=14
+      { card: c("spades", 13), playerIndex: 1 }, // K♠
+      { card: c("spades", 9), playerIndex: 2 },  // 9♠
+    ];
+    const state = mkState({
+      playerHands: [[], [], [], hand],
+      currentTrick: trick,
+      currentPlayerIndex: 3,
+    });
+    const pick = selectCardToPlay(hand, trick, state, 3, "medium");
+    expect(pick).toEqual(c("spades", 12));
+  });
+
+  it("does not dump Q♠ when last to play and would take the trick (no covering card)", () => {
+    // A♠ led and K♠ already played; Q♠ holder has Q♠ and would lose to A♠.
+    // But if only lower cards are covering Q♠ stays — wait, here A♠ covers so we DO dump.
+    // Variant: 9♠ is the current winner and Q♠ would win. Should NOT dump Q♠.
+    const hand = [c("spades", 12), c("spades", 7)];
+    const trick: TrickCard[] = [
+      { card: c("spades", 9), playerIndex: 0 },  // 9♠ led — winningRank=9
+      { card: c("spades", 3), playerIndex: 1 },
+      { card: c("spades", 5), playerIndex: 2 },
+    ];
+    const state = mkState({
+      playerHands: [[], [], [], hand],
+      currentTrick: trick,
+      currentPlayerIndex: 3,
+    });
+    const pick = selectCardToPlay(hand, trick, state, 3, "medium");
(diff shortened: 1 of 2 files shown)


wcmchenry3-stack:
## Post-fix simulator analysis

Ran the Hearts AI simulator before and after the Q♠ dump fix to validate the change.

### Setup
- **Pre-fix:** 100 games, `medium/medium/medium/medium`
- **Post-fix:** 500 games, `medium/medium/medium/medium`

### Key findings

| Metric | Pre-fix (100 games) | Post-fix (500 games) |
|--------|--------------------|--------------------|
| Max damage (Q♠ dumped on P3) | **25 pts** | **13 pts** |
| P3 avg pts/hand | 6.44 | 6.37 |
| Moon shots | 0/100 (0%) | 5/500 (1%) |
| Missed Q♠ dumps (0-pt bucket) | elevated | still present (see below) |

### Interpretation

The critical regression — where the AI was stuck holding Q♠ through multiple tricks, eventually absorbing 25 pts in a single hand — is **gone**. The post-fix max is 13 pts, consistent with unlucky-but-normal Hearts variance rather than a systematic bug.

P3 pts/hand improved slightly (6.44 → 6.37), which is the expected direction but modest given that Q♠-covering opportunities are relatively rare across 500 games × ~10 hands each.

Win distribution at 500 games is more balanced, with the flatter spread consistent with cleaner AI decision-making.

### Remaining question: 0-pt missed dumps

The analyzer still flags a number of "missed Q♠ dump" events in the 0-pt bucket. These may be:
1. **Analyzer false positives** — cases where the covering card was played *after* the Q♠ holder's position (the analyzer may not account for play order within the trick precisely)
2. **Non-last-position edge cases** — the fix targets `isLastToPlay` (trick.length === 3); seats 1 and 2 with a covering lead are a separate scenario not addressed here

This is worth a follow-up investigation but doesn't block this fix from shipping.

PR #1586, opened 2026-05-16
Hearts AI: prioritize A♣/K♣ in pass selection for Medium and Hard

### LINKED ISSUE #1584: Hearts AI: high clubs de-prioritized during pass selection, increasing early-trick risk
## Summary

The Medium and Hard AI pass strategies explicitly sort clubs to the **bottom** of the \"fill remaining\" candidate list, meaning A♣ and K♣ are the *last* cards an AI will pass when filling its third slot — even though high clubs are risky: 2♣ is always led on trick 1 and clubs get led heavily in the early tricks before hearts are broken, so holding A♣/K♣ forces the AI to win those early tricks with no way to offload points.

## Root Cause

In both `selectCardsToPassMedium` and `selectCardsToPassHard` (`frontend/src/game/hearts/ai.ts`), the \"fill remaining slots\" sort explicitly pushes clubs down:

```ts
// Medium — lines 149–151
if (a.suit === "clubs" && b.suit !== "clubs") return 1;
if (b.suit === "clubs" && a.suit !== "clubs") return -1;

// Hard — lines 233–236 (identical pattern)
if (a.suit === "clubs" && b.suit !== "clubs") return 1;
if (b.suit === "clubs" && a.suit !== "clubs") return -1;
```

This means A♣ (rank 1 → ace-high 14) is sorted *after* K♦ or K♥ during fill, despite being equally or more dangerous.

## Expected Behavior

High clubs (A♣, K♣, and possibly Q♣) should be treated as primary pass candidates — similar to how A♠/K♠ are passed when Q♠ is absent. The \"riskiest non-pointed card\" to hold going into a round is often a high club, not a high diamond, because:

1. 2♣ is mandatory trick-1 lead → clubs cycle through quickly in the early tricks
2. Holding A♣/K♣ guarantees winning those early club leads when hearts cannot be broken yet
3. Passing a high club is a low-cost pass (no points lost) with high risk-reduction value

## Suggested Fix

**Medium:** After the high-hearts and high-spades steps, add an explicit sub-step for A♣/K♣ (excluding 2♣ and clubs below 6) before the generic \"fill\" fallback, and remove the clubs de-prioritization from the sort.

**Hard:** Same treatment — elevate A♣/K♣ to an explicit priority tier (e.g., between step 3 high spades and step 4 void creation), and remove the anti-club sort bias.

**Easy:** No change needed — Easy already just passes the first 3 safe non-2♣ cards with no ordering logic.

## Verification

Run the simulator before and after the fix and compare average points per hand for player 0 across the Hard vs Hard batch:

```
npx tsx scripts/simulate-hearts.ts
```

A lower average score for the AI after the fix confirms the improvement. Alternatively use the hearts-analysis tool (`hearts-analysis/main.py`) with the `safe_discard_rate` endpoint to verify fewer wasted fill slots.

## Files

- `frontend/src/game/hearts/ai.ts` — `selectCardsToPassMedium` (line ~144) and `selectCardsToPassHard` (line ~228)
- `scripts/simulate-hearts.ts` — simulator for before/after comparison
- `hearts-analysis/analyzer.py` — `safe_discard_rate()` for deeper analysis

wcmchenry3-stack:
## Summary

Fixes #1584

High clubs (A♣, K♣) were being actively de-prioritized in the "fill remaining slots" sort in both Medium and Hard pass strategies, even though they're among the riskiest cards to hold going into a round — 2♣ is always led on trick 1 and clubs cycle heavily in early tricks before hearts are broken.

- **Medium:** Added explicit A♣/K♣ sub-step after the high-spades step, before the generic fill fallback. Removed the `suit === "clubs"` de-prioritization from the fill sort.
- **Hard:** Added explicit A♣/K♣ tier (step 3.5) between high spades and opportunistic void creation. Removed the same anti-club bias from the step-5 fill sort.

The existing `passSafeFilter` already excludes 2♣ and clubs below rank 6, so those constraints are respected naturally.

## Test plan

- [x] All 52 existing Hearts AI unit tests pass (`npx jest src/game/hearts/__tests__/ai.test.ts --no-coverage`)
- [ ] Run `npx tsx scripts/simulate-hearts.ts` before/after and confirm lower average points per hand for the AI

---
_Generated by [Claude Code](https://claude.ai/code/session_011fvK2xaqRZ1FcukKe5ooAh)_

### DIFF
--- a/frontend/src/game/hearts/__tests__/ai.test.ts
+++ b/frontend/src/game/hearts/__tests__/ai.test.ts
@@ -198,6 +198,55 @@ describe("selectCardsToPass", () => {
 });
 
 // ---------------------------------------------------------------------------
+// selectCardsToPass — Medium difficulty, high clubs (A♣/K♣)
+// ---------------------------------------------------------------------------
+
+describe("selectCardsToPass — Medium difficulty, high clubs", () => {
+  it("passes A♣ when slots remain after higher-priority cards", () => {
+    // Q♠ → slot 1, A♥ → slot 2, A♣ → slot 3 (step 3.5).
+    // Confirms rank 1 is not caught by the 'clubs below 6' guard (which checks rank > 1).
+    const hand = [
+      c("spades", 12),
+      c("hearts", 1),
+      c("clubs", 1),
+      c("spades", 3),
+      c("spades", 4),
+      c("spades", 5),
+      c("diamonds", 6),
+      c("diamonds", 7),
+      c("diamonds", 8),
+      c("clubs", 7),
+      c("clubs", 8),
+      c("hearts", 3),
+      c("hearts", 4),
+    ];
+    const passed = selectCardsToPass(hand, "left");
+    expect(passed).toContainEqual(c("clubs", 1));
+  });
+
+  it("passes K♣ when slots remain after higher-priority cards", () => {
+    // Q♠ → slot 1, A♥ → slot 2, K♣ → slot 3 (step 3.5).
+    const hand = [
+      c("spades", 12),
+      c("hearts", 1),
+      c("clubs", 13),
+      c("spades", 3),
+      c("spades", 4),
+      c("spades", 5),
+      c("diamonds", 6),
+      c("diamonds", 7),
+      c("diamonds", 8),
+      c("clubs", 7),
+      c("clubs", 8),
+      c("hearts", 3),
+      c("hearts", 4),
+    ];
+    const passed = selectCardsToPass(hand, "left");
+    expect(passed).toContainEqual(c("clubs", 13));
+  });
+});
+
+// ---------------------------------------------------------------------------
 // selectCardToPlay — void (discarding)
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


wcmchenry3-stack (commented):
## PR #1586 Review: `Hearts AI: prioritize A♣/K♣ in pass selection for Medium and Hard`

### Overview

Removes an active anti-club bias from the fill-sort fallback in both Medium and Hard pass strategies, and adds an explicit A♣/K♣ prioritization step before the generic fill. The motivation is sound: 2♣ is always the opening lead, so clubs cycle heavily in early tricks, making A♣/K♣ dangerous to hold.

---

### Correctness

**Anti-club bias removal** — Correct. The old sort pushed clubs to the back of the fill candidates (`return 1` if left card is clubs). Since A♣/K♣ are now explicitly handled in their own step, this de-prioritization was actively harmful and is right to remove.

**New A♣/K♣ step (Medium)** — Correct. None of steps 1–3 add clubs, so the `notSelected` guard isn't needed (and is correctly omitted). Both ranks `[1, 13]` (Ace, King) are tried in order.

**New A♣/K♣ step (Hard)** — Correct. Includes `notSelected(c)` defensively, consistent with the `notSelected` pattern used throughout the Hard function. Slots in cleanly between high spades (step 3) and opportunistic void creation (step 4).

**Priority ordering** — The strategy hierarchy makes sense: Q♠ → high hearts → high spades → high clubs → void creation → generic fill. Explicit dangerous cards take priority over strategic play.

**Void creation interaction** — Step 4 (Hard) fires only when `selected.length === 2`. If A♣/K♣ fill both remaining slots (1 → 3), void creation is correctly skipped. ✓

---

### Potential Issue

**`safe(c)` and A♣ (rank 1):** The PR says `passSafeFilter` excludes 2♣ and "clubs below rank 6." If the filter uses raw rank values, A♣ (rank `1`) would be `1 < 6` and incorrectly excluded — meaning the step only ever catches K♣, not A♣. The `aceHigh()` helper suggests rank 1 is treated as 14 elsewhere, but it's worth confirming `passSafeFilter` uses the same ace-high conversion (or explicitly whitelists rank 1). If A♣ is being filtered out, the A♣ bullet in the PR summary is misleading.

---

### Minor Observations

- **No new unit tests** for the specific A♣/K♣ prioritization. A test asserting that a hand containing A♣ or K♣ includes it in the pass selection (when slots remain) would prevent regression.
- **Simulation script unchecked** — the test plan has `[ ] Run simulate-hearts.ts` unmarked. This is the empirical validation that the fix actually helps. Should be run before merge.
- **Docs comment style** — the step numbering (`3.5`) in the JSDoc block is a pragmatic choice; it's consistent with how you wrote it in Medium. Acceptable.

---

### Code Quality

Clean and minimal. The change touches only what's needed. Comments explain the *why* (clubs are led early), not the *what*. The pattern is consistent between Medium and Hard.

---

### Verdict

**Approve with minor notes.** The logic is correct and the fix addresses a real strategic flaw. Two things before merge:

1. Verify `safe(c)` doesn't filter out A♣ (rank 1) due to the "clubs below rank 6" check.
2. Run `simulate-hearts.ts` and confirm it shows an improvement (or at worst no regression).

wcmchenry3-stack (commented):
Thanks for the thorough review!

**A♣ / `passSafeFilter` (potential issue):** Confirmed safe. The filter at line 86 reads `c.rank > 1 && c.rank < 6`, so A♣ (rank 1) is explicitly excluded from the "below 6" guard and passes through correctly. Only clubs 3–5 are blocked there, alongside the separate 2♣ check.

**New tests added (this commit):** Four new tests covering the code reviewer's regression concern:
- *Medium* — `passes A♣ when slots remain after higher-priority cards` (directly validates rank-1 filter behavior)
- *Medium* — `passes K♣ when slots remain after higher-priority cards`
- *Hard* — `passes A♣ before opportunistic void creation` (asserts step 3.5 fires before step 4, blocking void-creation from stealing the slot)
- *Hard* — `passes K♣ when slots remain after higher-priority cards`

All 56 tests pass.

Issue #1632, opened 2026-05-16
Simulator: add behavioral metrics for difficulty tier validation


wcmchenry3-stack:
## Context

As part of redesigning Hearts AI difficulty tiers (#1538), we defined concrete behavioral signatures for each difficulty level. The current simulator cannot validate whether the implementation meets these targets because it only tracks win rate, average score, hands per game, moon shots (any player), and Q♠ on human (binary, P0 only).

### 1. Per-seat Q♠ collection rate
The `queenOfSpades` event already has a `takerSeat` field — extend tracking to all 4 seats instead of just seat 0.

**Why:** Key Easy "tell" is that Easy takes Q♠ on itself in 25–35% of hands. Currently invisible in the simulator.

### 2. Void rate per seat per passing round
After `commitPass`, check each player's hand for missing suits. Track how often each player voided a suit.

**Why:** Hard should void a suit in >80% of passing rounds; Easy should be near 0%. Currently not tracked at all.

### 3. Moon shots by seat (who shot the moon)
The `moonShot` event already has a `shooter: number` field — use it. Report per-seat moon shot frequency and success rate.

**Why:** Hard should shoot the moon in ~10% of rounds (hands). Currently we only track a total count with no attribution. Note: a player either shoots the moon in a given round or they don't — this is a per-round binary, not a per-trick event.

### 4. Per-hand score distribution per player
Aggregate `handScores` across all hands for each seat. Report as mean ± std dev.

**Why:** Hard AI's own scores should follow a bimodal distribution (either controlled-low or high-risk); Easy scores should be more uniform and higher on average.

## Out of Scope

- Moon *attempt* rate (internal AI state, not exposed to game state — would require AI instrumentation)
- Q♠ routing decisions (inferred from Q♠-per-seat outcomes instead)

## Acceptance Criteria

- Simulator reports per-seat Q♠ collection rate for each batch
- Simulator reports per-seat void rate for each batch (passing rounds only)
- Simulator reports moon shot rate broken down by shooter seat
- Simulator reports per-player average hand score
- All existing batch results are still reported (no regressions in output format)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

wcmchenry3-stack:
## Baseline Results (before any difficulty tier changes)

```
Easy vs Easy vs Easy (baseline)
  Q♠/Hand by Seat:    24.9%  25.4%  24.9%  24.8%
[...]

Issue #1538, opened 2026-05-13
Enhancement: Hard Hearts AI difficulty not meaningfully harder than Medium


wcmchenry3-stack:
## Problem

The Hearts AI Hard difficulty is statistically indistinguishable from Medium. Simulation of 3,000 games per batch (Hard vs Medium + 2 Easy neutrals, seats swapped for mirror comparison) shows Hard at 52.5% win rate vs Medium at 52.5% — no significant difference (n.s.).

## Root Cause

Hard's intended differentiators fire too rarely or are not implemented:

### Moon-attempt mode
- Threshold: hold 8+ hearts + Q♠ + 0 points taken in current hand
- Estimated frequency: ~1–3% of hands
- Even when it fires, it rarely completes successfully — impact on aggregate win rate is negligible

### Endgame timing awareness
- Fires when any player reaches 65+ points
- By 65 points, outcomes are largely determined
- The Q♠-hold logic is defensive, not aggressive — it prevents handing someone else a win but doesn't create wins

### `chooseLeadHard` vs `chooseLead`
- These are **functionally identical**
- `seenKeys` is passed in but not used to actually change which suit to lead — both use the same longest-suit, lowest-card heuristic

### Void creation in passing
- Only triggers when `selected.length === 2` after high-priority cards (Q♠, high hearts, A♠/K♠) are exhausted
- This is rare since those high-priority cards usually fill all 3 pass slots

### High impact, low implementation cost
1. **Use seenKeys in lead decisions**: Lead the suit where dangerous high cards (A, K) have been played — forcing out remaining point cards. Currently seenKeys is built but ignored.
2. **Aggressive suit exhaustion in leading**: If A♠ and K♠ are both seen (gone), lead spades to force Q♠ out rather than avoiding them.

### Medium impact
3. **Widen void-creation in passing**: Change `selected.length === 2` to `selected.length <= 2` — void creation fires whenever a slot is available, not just the last slot.
4. **Lower moon-attempt threshold**: Reduce from 8 hearts to 7, increasing activation frequency while still being statistically rare.

### Longer-term
5. **Opponent score-weighted discarding**: Prefer dumping Q♠/high hearts onto the player currently winning (lowest score), not just anyone — this requires tracking scores during discard.

## Acceptance Criteria

- Simulation shows Hard achieving statistically significant edge over Medium in neutral-field test (batches 5 vs 6 in `simulate-hearts.ts`)
[...]

PR #1633, opened 2026-05-16
feat(hearts): Hard AI high-point trick conservation (#1626)

### LINKED ISSUE #1626: feat(hearts): Hard AI play-phase improvements to widen skill gap vs Medium
## Context

After implementing PRs #1604 (moon blocking, Q♠ lead fix, moon-attempt tracking) and #1624 (pass direction awareness), the simulator still shows Hard and Medium at essentially a coin flip in the neutral-field comparison (Hard 51.4% vs Medium 52.4%, n.s. over 18,000 games). Passing only happens 3 times per hand; the 13-trick play phase is where a meaningful Hard/Medium gap can be created.

See simulator results in epic #1538.

## Proposed improvements

These all live in `frontend/src/game/hearts/ai.ts`, specifically in `selectCardToPlayHard`.

### 1. Track whether Q♠ is still in play

Hard currently doesn't check if Q♠ has been played yet. When Q♠ is still unaccounted for (not in `wonCards` and not in the current trick), Hard should avoid winning spade tricks with A♠ or K♠. Being forced to win a spade trick while Q♠ is lurking is a common source of inadvertent 13-point pickups.

**Signal:** `!(state.wonCards.flat().some(isQueenOfSpades)) && !currentTrick.some(tc => isQueenOfSpades(tc.card))`

### 2. Exploit void suits more aggressively

When Hard is void in the led suit, it can discard freely. Current discard logic dumps the highest point card, which is good. But if Hard has been building a void on purpose (step 4 of passing), it should recognize these free-discard opportunities and prefer dumping Q♠ or high hearts rather than low safe cards.

**Current behavior:** Void discard picks highest point card or, if no point cards, lowest safe card.  
**Improved behavior:** Same, but explicitly prefer Q♠ > A♥ > K♥ > ... in the discard priority order (already partially true but not explicit).

### 3. Avoid winning multi-heart tricks when following suit

When following in the led suit, Hard currently plays "highest card that doesn't win the trick." It doesn't account for whether the trick has accumulated multiple hearts. If the trick already has 3+ hearts, winning it is 3+ points — Hard should be more willing to underplay even at the cost of not following optimal height.

**Change:** In the "following suit" branch, if `trickPoints(trick) >= 3` and Hard can play below the current winner, prefer the lowest card that still doesn't win rather than the highest non-winning card.

### 4. Smarter leading strategy

Both Medium and Hard currently lead the lowest card of the longest suit (via `chooseLeadHard`). Hard could be smarter:

- **Lead low toward a known short suit opponent:** If hearts are broken, leading hearts to force point-card pickups.
- **Lead second-longest suit** when the longest suit has no safe low cards.
- **Don't lead clubs once 2♣ has been played** if all remaining clubs are high (risks winning).

## Acceptance criteria

- [ ] Hard avoids winning unintended spade tricks when Q♠ is still live
- [ ] Hard correctly exploits void-suit leads for maximum point-card dumping
- [ ] Hard underplays in high-point-value tricks to avoid taking 3+ hearts at once
- [ ] Simulator re-run shows Hard win rate > Medium win rate in the neutral-field batch at p < 0.05
- [ ] All existing AI tests pass; new tests cover each new behavior
- [ ] `npx prettier --write` passes before commit

## Notes

- All AI code is in `frontend/src/game/hearts/ai.ts` (TypeScript client-side), not `backend/hearts/module.py`
- Simulator: `npx tsx scripts/simulate-hearts.ts` — 6 batches × 3,000 games (~27s)
- Target functions: `selectCardToPlayHard`, `chooseLeadHard`

wcmchenry3-stack:
## Summary

- Adds high-point trick conservation to `selectCardToPlayHard`: when a trick already holds ≥3 hearts, plays the lowest losing in-suit card instead of the highest, preserving high cards for future use
- Two other play-phase improvements (Q♠-live spade guard, heart flushing) were implemented and reverted after simulation showed board control regression — losing the lead lets opponents exploit strategic position more than these guards helped

## Test plan

- [ ] 76 Hearts AI unit tests pass (`cd frontend && npx jest --testPathPattern="hearts/.*ai"`)
- [ ] Two new Golden Hands tests cover the high-point conservation behavior (`#1626 high-point trick conservation (Hard)`)

Closes #1626


wcmchenry3-stack:
Closing in favor of the reframed difficulty tier work (#1538). The high-point trick conservation is logically correct but doesn't address the behavioral gaps identified by the new simulator metrics (#1632) — void rate (5% vs 80% target), moon shot rate (0.1% vs 10% target), and adversarial targeting (not implemented). New issues will be scoped to those behavioral targets instead.

PR #1684, opened 2026-05-17
fix(mahjong): remove rubber-band bounce on zoom limit

### LINKED ISSUE #1683: Mahjong: remove rubber-band bounce when zooming past max
## Problem

When pinching to zoom in on the Mahjong board past the maximum zoom level, the view overshoots and then spring-animates back to the limit. This bounce is disorienting — the user would prefer the zoom to simply stop at the ceiling with no rebound.

## Root cause

Two things in `frontend/src/screens/MahjongScreen.tsx` produce the bounce:

1. **`rubberClamp` in `onUpdate`** (`zoom.ts:56-61`, used at `MahjongScreen.tsx:361`) — applies a 30% resistance factor (`RUBBER_FACTOR = 0.3`) that lets `zoomScale` exceed `maxZoom` while the pinch is active.
2. **`withSpring` in `onEnd`** (`MahjongScreen.tsx:367`) — when the gesture ends with `zoomScale > maxZoom`, it animates back using a spring (damping 20, stiffness 300), producing the visible bounce.

## Fix

- In `pinchGesture.onUpdate`, replace `rubberClamp(...)` with the existing hard `clamp(...)` so `zoomScale` never exceeds bounds during the gesture.
- In `pinchGesture.onEnd`, remove the `withSpring` path — since `zoomScale` will already be within bounds, no snap-back is needed. `baseScale` can be set directly.
- Remove `rubberClamp` and `RUBBER_FACTOR` from `zoom.ts` (now unused).

## Acceptance

- Pinching in past max zoom stops immediately at `maxZoom` with no bounce or spring animation.
- Pinching out past min zoom stops immediately at `minZoom` with no bounce.
- Normal zoom-in/out within bounds still feels smooth.
- Existing `zoom.ts` unit tests continue to pass (after removing the `rubberClamp` tests).

wcmchenry3-stack:
$(cat <<'EOF'
## Summary

- Replaces `rubberClamp` with hard `clamp` in the pinch `onUpdate` handler so `zoomScale` never exceeds `[minZoom, maxZoom]` during the gesture
- Removes the `withSpring` snap-back in `onEnd` — no overshoot means no correction needed
- Deletes the now-unused `rubberClamp` function, `RUBBER_FACTOR` constant, and their unit tests

Closes #1683

## Test plan

- [ ] Pinch to zoom in past max — zoom stops immediately at ceiling, no bounce
- [ ] Pinch to zoom out past min — zoom stops immediately at floor, no bounce
- [ ] Normal zoom within bounds still feels smooth
- [ ] All 15 `zoom.test.ts` unit tests pass
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_019ecjjXJPFuGJf1unc116Uj)_

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -8,7 +8,6 @@
       "name": "frontend",
       "version": "1.0.8",
       "dependencies": {
-        "@dimforge/rapier2d-compat": "^0.19.3",
         "@expo-google-fonts/manrope": "^0.4.2",
         "@expo-google-fonts/space-grotesk": "^0.4.1",
         "@expo/vector-icons": "^15.1.1",
@@ -1488,12 +1487,6 @@
       "integrity": "sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==",
       "dev": true,
       "license": "MIT"
-    },
-    "node_modules/@dimforge/rapier2d-compat": {
-      "version": "0.19.3",
-      "resolved": "https://registry.npmjs.org/@dimforge/rapier2d-compat/-/rapier2d-compat-0.19.3.tgz",
-      "integrity": "sha512-5HZUgiSpu9Uba3He+4SCPVte3h2vgxcodkxWThRjd3X/zgq/Z4OyRUKpNEOsOR6AC+OfulPkkZ5asZGzRMpcwA==",
-      "license": "Apache-2.0"
     },
     "node_modules/@egjs/hammerjs": {
       "version": "2.0.17",
--- a/frontend/src/game/mahjong/__tests__/zoom.test.ts
+++ b/frontend/src/game/mahjong/__tests__/zoom.test.ts
@@ -4,8 +4,6 @@ import {
   computeZoomBounds,
   MIN_READABLE_TILE_PX,
   MIN_ZOOM_HEADROOM,
-  rubberClamp,
-  RUBBER_FACTOR,
 } from "../zoom";
 
 describe("computeZoomBounds", () => {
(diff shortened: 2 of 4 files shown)


wcmchenry3-stack:
## Code Review

**3 additions / 57 deletions** — a net shrink, which is a good sign for a fix.

---

### Overview

Replaces the rubber-band zoom behavior with a hard clamp, then removes all the now-dead code (`rubberClamp`, `RUBBER_FACTOR`, their tests, and the `withSpring` snap-back). The behavioral goal is clear and the implementation follows logically.

---

### Code Quality

**MahjongScreen.tsx (the actual fix)**
- `onUpdate` change is correct: `clamp` instead of `rubberClamp` means `zoomScale` never leaves `[minZoom, maxZoom]` during the gesture.
- `onEnd` simplification is valid: because `zoomScale` is always in-bounds, `target === zoomScale.value` always, so the old `Math.abs` guard and `withSpring` branch were dead code. The new one-liner `baseScale.value = zoomScale.value` is clean.
- `withSpring` may now be an unused import — worth verifying there's no TS/lint warning (it might be used elsewhere in the screen, but the diff removes its only visible usage).

**zoom.ts / zoom.test.ts**
- Dead code removal is thorough and correct. No orphaned references remain.

---

### Concerns

**`package-lock.json` — `@dimforge/rapier2d-compat` removed**

This is the biggest flag. The diff removes `@dimforge/rapier2d-compat` from the `dependencies` block (which mirrors `package.json`) and strips its `node_modules` entry entirely. This is unrelated to Mahjong zoom and is not mentioned in the PR description. Given the active `feat/1610-cascade-uc1-angular-damping` work, Cascade appears to rely on Rapier2D for physics. If this removal was a rebase artifact or accidental `npm install` side-effect, merging it would break Cascade builds.

**Action required:** Confirm whether `@dimforge/rapier2d-compat` was intentionally removed from `frontend/package.json`. If not, revert that hunk before merging.

---

### Test Coverage

- 5 `rubberClamp` tests correctly deleted alongside the function.
- Remaining 15 `zoom.test.ts` tests cover `computeZoomBounds`, `computePanBounds`, and `clamp` — still valid.
- No new tests needed; the replacement behavior (`clamp`) is already tested by the existing `clamp` suite.
- Manual test plan in the PR description is appropriate for the gesture path.

---

### Summary

The Mahjong zoom fix itself is clean, minimal, and correct — good PR. The one blocker is the `@dimforge/rapier2d-compat` removal in `package-lock.json`. Verify that's intentional (and update `package.json` accordingly if so) or revert the hunk before merging.

---
_Reviewed with [Claude Code](https://claude.ai/code)_

wcmchenry3-stack:
**Follow-up:** Both concerns from the previous review are non-issues on closer inspection.

- **`@dimforge/rapier2d-compat`** — already absent from `dev`'s `package.json`; commit `e7dd2a08` on this branch explicitly syncs `package-lock.json` to match. The removal was intentional and pre-dates this PR.
- **`withSpring` import** — still used at lines 232 and 962 of `MahjongScreen.tsx`; not orphaned.

PR is clear to merge. 🟢

PR #1723, opened 2026-05-19
fix(assets): remove duplicate Mahjong tile asset load on web

### LINKED ISSUE #1629: fix(assets): remove duplicate Mahjong tile asset load on web
## Bug

Both `GameCanvas.web.tsx` and `MahjongScreen.tsx` independently call `Asset.fromModule().downloadAsync()` on all 42 SVG tiles when the Mahjong screen mounts. This means the tiles are loaded and decoded twice on web, doubling the asset initialization work on every game entry.

## Root cause

Two separate components each run a full `Asset.fromModule().downloadAsync()` loop over the tile set on mount, with no shared cache guard between them.

## Fix

- Determine the correct single load point for the tiles (likely `MahjongScreen.tsx` as the screen-level owner)
- Remove the redundant call from `GameCanvas.web.tsx` (or vice versa, depending on which owns the asset lifecycle)
- Ensure the remaining load path still works correctly before the canvas renders

## Key files
- `frontend/src/games/mahjong/GameCanvas.web.tsx`
- `frontend/src/games/mahjong/MahjongScreen.tsx`

## Acceptance criteria
- [ ] Mahjong tiles load exactly once on web (verify via browser Network tab — 42 SVG requests, not 84)
- [ ] Tiles render correctly after the fix
- [ ] No regressions on native (native path is separate)

wcmchenry3-stack:
## Summary

Closes #1629

- Both `GameCanvas.web.tsx` and `MahjongScreen.tsx` independently called `Asset.fromModule().downloadAsync()` over all 42 SVG tiles on mount, causing 84 network requests instead of 42.
- Introduce `tileAssetLoader.web.ts` with a module-level singleton promise so the first caller downloads all tiles and every subsequent caller (including concurrent ones) awaits the same in-flight result.
- A native stub (`tileAssetLoader.ts`) is provided for Metro platform resolution — the native code path is unchanged.

## Changes

- **`frontend/src/components/mahjong/tileAssetLoader.web.ts`** (new) — singleton `loadTileAssets()` that downloads once and caches the promise
- **`frontend/src/components/mahjong/tileAssetLoader.ts`** (new) — native stub returning `Promise.resolve([])`
- **`frontend/src/components/mahjong/GameCanvas.web.tsx`** — calls `loadTileAssets()` to get pre-downloaded URIs, then creates `HTMLImageElement` objects; drops direct `expo-asset` usage
- **`frontend/src/screens/MahjongScreen.tsx`** — calls `loadTileAssets()` for `FlyingPair` overlay URIs; drops direct `expo-asset` usage

## Test plan

- [ ] Open Mahjong on web and check the browser Network tab — 42 SVG requests, not 84
- [ ] Tiles render correctly on the board
- [ ] Matched-pair FlyingPair animation shows correct tile art
- [ ] No regressions on native (native path is separate, stub is a no-op)

https://claude.ai/code/session_01HozzRZh7KuwPVWuZR4Bz1W

---
_Generated by [Claude Code](https://claude.ai/code/session_01HozzRZh7KuwPVWuZR4Bz1W)_

### DIFF
--- a/frontend/src/components/mahjong/GameCanvas.web.tsx
+++ b/frontend/src/components/mahjong/GameCanvas.web.tsx
@@ -9,11 +9,10 @@
 
 import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { Pressable, StyleSheet, Text, View } from "react-native";
-import { Asset } from "expo-asset";
 import { useTranslation } from "react-i18next";
 import { getMatchingFreeTileIds, hasFreePairs, isFreeTile } from "../../game/mahjong/engine";
 import type { MahjongState, SlotTile } from "../../game/mahjong/types";
-import { TILE_REQUIRES } from "./tileAssets";
+import { loadTileAssets } from "./tileAssetLoader";
 import {
   MAHJONG_BOARD_BG,
   MAHJONG_GLOW_SHADOW,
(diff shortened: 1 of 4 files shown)


wcmchenry3-stack (commented):
## Code Review

### Overview

Fixes a double-load bug (#1629) where `GameCanvas.web.tsx` and `MahjongScreen.tsx` independently called `Asset.fromModule().downloadAsync()` for all 42 SVG tiles on mount, producing 84 network requests. The solution introduces a module-level singleton promise (`tileAssetLoader.web.ts`) so only the first caller downloads and all subsequent callers await the same in-flight result. Metro platform resolution handles the native/web split cleanly.

---

### Issues

**Dead import in `MahjongScreen.tsx`** (actionable)

`TILE_REQUIRES` is still imported but is no longer used after the refactor — its only consumer was `Array(TILE_REQUIRES.length).fill(null)`, which was removed:

\`\`\`ts
// still present after the PR — should be removed
import { TILE_REQUIRES } from "../components/mahjong/tileAssets";
\`\`\`

**Unchained inner `Promise.all` in `GameCanvas.web.tsx`** (minor)

The inner `Promise.all` inside the `.then()` callback is not returned or caught:

\`\`\`ts
loadTileAssets().then((uris) => {
  if (cancelled) return;
  Promise.all(   // ← not returned; rejection (however unlikely) silently swallows
    uris.map(...)
  ).then(() => { ... });
});
\`\`\`

In practice this can't reject (all errors are caught via `onerror`), but the pattern is fragile. A simple fix: `return Promise.all(...)`.

---

### Positives

- **Singleton is correct and race-safe.** JavaScript's single-threaded model means `if (!_promise) { _promise = ... }` is atomic — no TOCTOU risk.
- **Quiet URI fix.** The old `MahjongScreen.tsx` used `asset.uri ?? null`; the new loader consistently uses `asset.localUri ?? asset.uri ?? null`, which is the correct preference order for cached assets.
- **`readonly` return type** on `loadTileAssets()` prevents callers from mutating the cached result.
- **Metro platform split** (`tileAssetLoader.web.ts` / `tileAssetLoader.ts`) is the idiomatic approach; native path is untouched.
- **`cancelled` flag** is correctly checked in both callers before applying async results to state.

---

### Summary

Clean, well-scoped fix. Two items before merge: remove the `TILE_REQUIRES` dead import from `MahjongScreen.tsx`, and optionally chain/return the inner `Promise.all` in `GameCanvas.web.tsx` for correctness hygiene.

Issue #1565, opened 2026-05-15
bug(mahjong): shuffle can create unwinnable state by stacking a matching pair


wcmchenry3-stack:
## Summary

When a shuffle is triggered near the end of a game, the shuffle algorithm can place the two tiles of a matching pair directly on top of each other at the same `(col, row)` grid position. The result is permanently unwinnable: the top tile is free and can be selected, but removing it leaves the bottom tile with no partner — the game is stuck even though a shuffle was consumed.

The screenshot below shows the symptom: **NO MOVES** appears with 71/72 pairs removed and 1 shuffle remaining, but the shuffle produces the same deadlocked layout.

## Root Cause

`shuffleBoard` in `frontend/src/game/mahjong/engine.ts` retries until `hasFreePairs` returns `true` (at least one accessible matching pair exists). However, it does **not** check whether a pair's two tiles are placed at the same `(col, row)` — only differing by `layer`. In that configuration:

- The top tile (`layer + 1`) is free (nothing above it, at least one open side).
- The bottom tile (`layer`) is blocked by the top tile and is never accessible.
- Once the top tile is removed, the bottom tile's partner is gone → unrecoverable deadlock.

With only 2 tiles remaining, the two slot positions may both be within the same column stack, so roughly half of all shuffle attempts produce this exact problem.

## How Other Games Handle It

| Game | Approach |
|---|---|
| Microsoft Mahjong / Mahjong Titans | Deal-time constraint: matching pairs are never assigned to (col, row)-sharing stacks |
| iOS "Mahjong Solitaire Epic" | Full forward-reachability solver after every shuffle (feasible for small tile counts) |
| Most web Mahjong apps | Simple rule: no two matching tiles may share the same `(col, row)` |

## Fix (already implemented on `claude/fix-overlapping-pieces-fkCCw`)

Added a rejection check inside the shuffle retry loop. After each candidate assignment, the loop now verifies that no pair's two slots share `(col, row)`. If they do, the candidate is discarded and a new random assignment is tried.

```typescript
const hasStackedMatch = shuffledPairs.some((_, i) => {
  const sA = shuffledSlots[i * 2]!;
  const sB = shuffledSlots[i * 2 + 1]!;
  return sA.col === sB.col && sA.row === sB.row;
});
if (!hasStackedMatch && hasFreePairs(candidate)) {
  newTiles = candidate;
  break;
}
```

[...]

PR #1728, opened 2026-05-19
test(mahjong): end-game shuffle regression tests for issue 1565 stacked-pair deadlock

wcmchenry3-stack:
$(cat <<'EOF'
## Summary

- Adds three acceptance-criteria tests to the `shuffleBoard` suite covering the stacked-pair deadlock bug from issue #1565
- The engine fix (`hasStackedMatch` guard in `shuffleBoard`) was already merged; these tests lock in regression coverage

## What's tested

| Test | Scenario | Seeds |
|---|---|---|
| 1 pair remaining always winnable | Minimal 2-tile state at non-stacked slots | 30 |
| Issue 1565 deadlock reproduction | 2-pair state with a column stack — the exact bad-shuffle that `hasStackedMatch` blocks | 50 |
| 2–4 pairs end-game stress | Seeded game played down to 8 tiles, shuffled | 50 |

All three verify that after `shuffleBoard`, no matching pair's two tiles share `(col, row)` and `hasFreePairs` remains true.

## Test plan

- [x] `npm test -- --testPathPattern="mahjong/__tests__/engine"` → 82 tests, all green
- [x] No changes to engine logic — pure test additions

https://claude.ai/code/session_011EGsHjfBgkBghDbinTZeDr
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_011EGsHjfBgkBghDbinTZeDr)_

### DIFF
--- a/frontend/src/game/mahjong/__tests__/engine.test.ts
+++ b/frontend/src/game/mahjong/__tests__/engine.test.ts
@@ -563,9 +563,18 @@ describe("shuffleBoard", () => {
     const a: SlotTile = { id: 0, suit: "characters", rank: 1, faceId: 8, col: 0, row: 0, layer: 0 };
     const b: SlotTile = { id: 1, suit: "characters", rank: 1, faceId: 8, col: 2, row: 0, layer: 0 };
     const state: MahjongState = {
-      _v: 1, tiles: [a, b], pairsRemoved: 71, score: 710, shufflesLeft: 1,
-      selected: null, undoStack: [], isComplete: false, isDeadlocked: false,
-      startedAt: null, accumulatedMs: 0, dealId: "TEST",
+      _v: 1,
+      tiles: [a, b],
+      pairsRemoved: 71,
+      score: 710,
+      shufflesLeft: 1,
+      selected: null,
+      undoStack: [],
+      isComplete: false,
+      isDeadlocked: false,
+      startedAt: null,
+      accumulatedMs: 0,
+      dealId: "TEST",
     };
     for (let seed = 0; seed < 50; seed++) {
       setRng(createSeededRng(seed));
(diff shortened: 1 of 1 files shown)


wcmchenry3-stack (commented):
## PR #1728 Review — `test(mahjong): end-game shuffle regression tests for issue 1565`

### Overview

Pure test addition (80 lines, no logic changes) locking in regression coverage for the stacked-pair deadlock bug fixed in issue #1565. The `hasStackedMatch` guard in `shuffleBoard` already prevents the bad shuffle; these tests verify it doesn't regress. Three scenarios: trivial 1-pair case, exact bug reproduction, and an end-game stress sweep.

---

### Code Quality

The tests are clearly named, well-commented, and the slot-layout setup in test 2 accurately models the original deadlock scenario. The bug description in the comment is precise enough that a future reader could reproduce the issue without reading the original ticket.

---

### Issues

**1. Silent false-negatives when all shuffle retries fail (tests 2 & 3)**

Both tests have:
\`\`\`typescript
if (shuffled === state) continue; // all 50 retries failed (should not happen)
\`\`\`

The comment is correct — this path represents the engine returning the unmodified state because every attempt was rejected. But silently \`continue\`-ing past the \`expect()\` calls means if the guard is ever accidentally removed and all seeds now hit a stacked layout, the test passes vacuously. This path should be an explicit failure:

\`\`\`typescript
expect(shuffled).not.toBe(state); // guard must find a valid arrangement
\`\`\`

**2. Test 3 missing `hasFreePairs` assertion**

Tests 1 and 2 both assert \`hasFreePairs(shuffled.tiles) === true\`. Test 3 only checks the no-stacked-pair condition. For consistency — and to catch a hypothetical regression where \`hasStackedMatch\` passes but \`hasFreePairs\` somehow fails — add:

\`\`\`typescript
expect(hasFreePairs(shuffled.tiles)).toBe(true);
\`\`\`

**3. Test 1 provides weak regression coverage for the guard**

With 2 tiles at \`(col:0, row:0)\` and \`(col:2, row:0)\`, there are only 2 slots and neither shares \`(col, row)\`. A shuffle of these two slots can never produce a stacked pair — the \`hasStackedMatch\` guard is unreachable. The test will pass even if the guard is deleted. It still serves as a sanity check, but its regression value for this specific fix is low. Consider noting this in the comment, or replacing it with a scenario where stacking is geometrically possible for 1 pair.

---

### Minor Notes

- \`faceId: 8\` / \`faceId: 1\` in test 2 are arbitrary (correct — \`tilesMatch\` ignores \`faceId\`), but a quick inline note like \`// faceId irrelevant to match logic\` would pre-empt the "why 8?" question.
- The 30-seed count in test 1 vs 50 in tests 2 & 3 is unexplained. Not a bug, but worth a brief rationale if the counts were intentional.

---

### Test Coverage Assessment

| Scenario | Correctness | Regression Value |
|---|---|---|
| 1 pair, non-stacked | ✓ | Low (guard unreachable) |
| Issue 1565 exact repro | ✓ | High (correct reproduction) |
| 2–4 pair stress | ✓ | Medium (missing `hasFreePairs`) |

---

### Verdict

**Approve with minor fixes requested.** The bug reproduction (test 2) is accurate and the overall structure is solid. The two actionable issues are: replace the silent \`continue\` with an \`expect(...).not.toBe(state)\` in tests 2 and 3, and add a \`hasFreePairs\` assertion to test 3.

PR #1742, opened 2026-05-19
fix: move CORSMiddleware outside rate limiter so 429/413 errors carry CORS headers

### LINKED ISSUE #1739: bug: GET /entitlements fails with network error on web platform (production)
## Summary

`GET /entitlements` is failing with a `Failed to fetch` network error on the web platform in production. Sentry reports this as an ongoing issue: **342 events across 15 users**.

Sentry issue ID: `3ae15d2c` · Level: Warning · Ongoing

## Observed behaviour

The API client breadcrumbs show:

```
GET: https://dev-games-api.buffingchi.com/entitlements → Failed to fetch
BASE_URL=https://dev-games-api.buffingchi.com, platform=web
```

The `EntitlementContext` catches the `TypeError` and logs a Sentry warning rather than crashing, so users are silently left without a refreshed entitlement token. The 7-day offline grace period prevents immediate access loss, but repeated failures will eventually downgrade users.

## Affected environment

- **Environment:** production
- **Release:** 1.0.9
- **Platform:** web (Expo Web)
- **Device sample:** Motorola g power (2022), Android 11, Chrome Mobile 119

## Likely cause

The `/entitlements` endpoint on `dev-games-api.buffingchi.com` may not include CORS response headers for web platform requests. Native (iOS/Android) requests go through the system network stack and are not subject to CORS; web requests are. If `Access-Control-Allow-Origin` is missing or scoped too narrowly on that route, the browser silently drops the response and the client receives a generic network error.

## Steps to reproduce

1. Build and serve the Expo web bundle against the production API.
2. Open in a browser (Chrome, Chrome Mobile).
3. Observe Sentry warning: `API entitlements network failure: GET /entitlements`.

## Acceptance criteria

- [ ] `/entitlements` responds with correct CORS headers for web origins (or the failure mode is confirmed to be something else and the real root cause is addressed).
- [ ] No new Sentry warnings of type `API entitlements network failure` on web platform.
- [ ] Existing users are not inadvertently downgraded during the fix.

## References

- `frontend/src/entitlements/EntitlementContext.tsx` — fetch + error handling
- `frontend/src/game/_shared/httpClient.ts` — `resolveBaseUrl()`, CORS-adjacent breadcrumb logging

wcmchenry3-stack:
$(cat <<'EOF'
## Summary

- `CORSMiddleware` was registered innermost (first `add_middleware` call), placing it *inside* `SlowAPIMiddleware` and `MaxBodySizeMiddleware`. Early error responses (429 rate-limit, 413 body-too-large) bypassed CORS entirely — no `Access-Control-Allow-Origin` header — so browsers raised `TypeError: Failed to fetch`, silently leaving web users without a refreshed entitlement token (342 Sentry events, 15 users, #1739).
- Reordered `add_middleware` calls so `CORSMiddleware` is registered last (outermost of those three), ensuring all responses pass through CORS.
- Fixed `@app.middleware` definition order (`security_headers` → `request_logger`) so `request_logger` is genuinely the outermost middleware as the existing comment always intended.
- Added 3 CORS tests: allowed-origin GET, OPTIONS preflight, disallowed-origin exclusion.

## Test plan

- [ ] All 569 existing tests pass (verified locally)
- [ ] `test_cors_get_entitlements_includes_allow_origin` — GET with valid `Origin` returns `Access-Control-Allow-Origin`
- [ ] `test_cors_preflight_entitlements` — OPTIONS preflight returns CORS headers with `GET` in allowed methods
- [ ] `test_cors_disallowed_origin_excluded` — unlisted origin receives no `Access-Control-Allow-Origin` header
- [ ] Deploy to dev and confirm no new `API entitlements network failure` Sentry warnings on web platform

Closes #1739

https://claude.ai/code/session_01NNPZ3K24ZoWpP1wmkZmLi9
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_01NNPZ3K24ZoWpP1wmkZmLi9)_

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -128,11 +128,11 @@ _allowed_origins: list[str] = (
 # ---------------------------------------------------------------------------
 # Middleware stack (registered last = outermost in Starlette)
 # Order outermost → innermost:
-#   1. request_logger  (logs all requests including 429s)
-#   2. SlowAPIMiddleware  (rate limiting)
-#   3. MaxBodySizeMiddleware  (reject oversized bodies early)
-#   4. CORSMiddleware
-#   5. security_headers  (@app.middleware decorator, innermost)
+#   1. request_logger  (@app.middleware, outermost — logs every request incl. 429s)
+#   2. security_headers  (@app.middleware)
+#   3. CORSMiddleware  (must wrap SlowAPI/MaxBody so 429/413 errors carry CORS headers)
+#   4. SlowAPIMiddleware  (rate limiting)
+#   5. MaxBodySizeMiddleware  (reject oversized bodies early, innermost)
 # ---------------------------------------------------------------------------
 
 DEFAULT_MAX_BODY_BYTES = 1_024  # 1 KB — legacy game payloads (~50 bytes max)
@@ -161,7 +161,13 @@ class MaxBodySizeMiddleware(BaseHTTPMiddleware):
         return await call_next(request)
 
 
-# Register in reverse of desired execution order (last registered = outermost)
+# Register in reverse of desired execution order (last registered = outermost).
+# CORSMiddleware is registered last so it wraps SlowAPI and MaxBody — this
+# ensures 429 (rate-limited) and 413 (body-too-large) responses include the
+# Access-Control-Allow-Origin header. Without it, browsers block those error
+# responses and raise TypeError: Failed to fetch (#1739).
+app.add_middleware(MaxBodySizeMiddleware)
+app.add_middleware(SlowAPIMiddleware)
 app.add_middleware(
     CORSMiddleware,
     allow_origins=_allowed_origins,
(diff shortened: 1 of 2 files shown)


wcmchenry3-stack:
## Code Review

### Overview

Correct fix for a real user-facing bug (342 Sentry events, 15 users). `CORSMiddleware` was registered innermost, so `SlowAPIMiddleware` (429) and `MaxBodySizeMiddleware` (413) short-circuit responses were never wrapped by CORS — browsers raised `TypeError: Failed to fetch` and swallowed the error silently.

The reordering is correct: last-registered = outermost in Starlette, so registering `CORSMiddleware` last of those three makes it the outer wrapper. The `@app.middleware` definition order fix for `security_headers` → `request_logger` is also correct.

---

### Main Issue: Missing regression test for the actual bug scenario

The three new tests verify CORS on `200 GET` and `OPTIONS` preflight, but the **specific failure path — a 429 or 413 response including `Access-Control-Allow-Origin`** — is not covered. If someone swaps the `add_middleware` order again, no test catches it.

At minimum, add the 413 case (easy to trigger without timing concerns):

```python
def test_cors_oversized_body_response_includes_allow_origin(client: TestClient) -> None:
    """413 responses must carry CORS headers so browsers can read them (#1739)."""
    r = client.post(
        "/scores",
        content=b"x" * (DEFAULT_MAX_BODY_BYTES + 1),
        headers={"Origin": _TEST_ORIGIN, ...},
    )
    assert r.status_code == 413
    assert r.headers.get("access-control-allow-origin") == _TEST_ORIGIN
```

If hitting the rate limit in tests is feasible, a 429 version of the same test would fully close the regression loop.

---

### Minor Notes

- **CSP still references dev URLs** (`dev-games-api.buffingchi.com`, `dev-games.buffingchi.com`) — pre-existing, but worth a follow-up if this goes to prod as-is.
- `expose_headers=["Retry-After"]` already on `CORSMiddleware` means browsers can read `Retry-After` on 429s once CORS is delivered. Good.
- The inline "why" comment for the `CORSMiddleware` position is exactly the right level of detail — keep it.

---

**Verdict:** Approve with the suggestion to add at least the 413+CORS regression test before merging.

Issue #1803, opened 2026-05-22
Bug: Bottle sort pour animation breaks on cross-row moves (level 15+)


wcmchenry3-stack:
## Summary

Pour animations visually break starting at level 15. The ghost bottle tilts in the wrong direction and the liquid stream appears on the wrong side of the bottle when pouring between bottles that are in different rows and whose visual left-right order is opposite their index order.

## Root Cause

**File:** `frontend/src/game/sort/components/SortBoard.tsx`, line 205

```ts
// Current (broken for multi-row layouts)
const isRight = pouringFrom < pouringTo;
```

`isRight` determines:
1. Which side of the bottle to place the spout pivot (`NECK_RIGHT_VB` vs `NECK_LEFT_VB`)
2. The tilt direction of the ghost animation

This works when all bottles are in a single row (levels 1–6) because lower index = leftward. But level 15 has **9 bottles in a 4-4-1 grid (3 rows)**:

```
Row 1: [0] [1] [2] [3]
Row 2: [4] [5] [6] [7]
Row 3:         [8]
```

For a pour from bottle **3** (rightmost of row 1) to bottle **4** (leftmost of row 2):
- Index check: `3 < 4` → `isRight = true` → ghost tilts **right**, spout on right side
- Visually: bottle 4 is to the **left** of bottle 3 → should tilt **left**

The ghost ends up tilting away from the destination and the stream appears on the opposite side from where the liquid is going.

## Fix

Replace the index-based check with a visual x-position comparison (positions are already computed before this line):

```ts
// Fixed: compare actual visual positions
const isRight = srcPos.x < dstPos.x;
```

`srcPos` and `dstPos` are already available at line 197–198 from `bottlePositionsRef`.

## Affected Levels

Any level where the grid wraps into multiple rows **and** a pour crosses row boundaries with a reversed visual order. Level 15 is the first level with a 3-row layout (9 bottles), making it the first reliably affected level.

## Reproduction

1. Start level 15
2. Tap bottle at index 3 (top row, rightmost) to select it
3. Tap bottle at index 4 (middle row, leftmost) as destination
4. Observe: ghost bottle tilts right and stream appears on the wrong side

PR #1806, opened 2026-05-22
fix: Bottle sort pour animation breaks on cross-row moves (issue #1803)

vanhci:
Fixes issue #1803 - Bottle sort pour animation breaks on cross-row moves (level 15+)

## Problem
In SortBoard.tsx line 205, `isRight = pouringFrom < pouringTo` uses index-based comparison. For multi-row layouts (7+ bottles with 4 columns), visual left-right order differs from index order. For example, bottle 4 (row 2, col 0) is visually LEFT of bottle 3 (row 1, col 3), but `4 < 3` is false, causing the pour animation to tilt the wrong direction.

## Fix
Changed `isRight` to use actual visual x-position comparison: `const isRight = srcPos.x < dstPos.x;`

This correctly determines the pour direction based on where bottles are visually positioned in the grid, not their index.


wcmchenry3-stack:
Closing in favor of #1812, which addresses all review findings:

- **Retargeted to `dev`** (this PR targeted `main`, violating the project's git workflow)
- **Scoped to 2 files** — the 386-file diff was noise from the `main`-vs-`dev` delta; the actual change is `SortBoard.tsx` + its test
- **Fixed second bug site** — `pouringDirection` (reduce-motion fallback path, line 316) had the same `pouringFrom < pouringTo` index comparison and is fixed in the new PR
- **Added regression test** — `renders cross-row pour without crashing (regression #1803)`
- **ESLint + Prettier clean**

The bundle-limit change (`5.0 MB → 6.0 MB` in `ci.yml`) and all backend sort changes are already in `dev` from prior merged PRs, so they don't appear in the new diff.

wcmchenry3-stack:
Superseded by #1812.

PR #1831, opened 2026-05-27
Release: dev → main (conflicts resolved)

wcmchenry3-stack:
## Summary

All 171 commits from dev merged into main. 40 merge conflicts resolved — all in favor of dev (the newer code).

- **Cascade v2** engine rewrite (Matter.js physics, combos, persistence, piece queue)
- **Maestro** E2E smoke suite for all 10 games + CI jobs with reporting
- **Yacht** AI turn animation slowdown + end-game scorecard with AI scores
- **Sort** animation fixes, level progression, async BFS, cross-row pour fix
- **Mahjong** testIDs, shuffle/engine test coverage
- **StarSwarm** Free Fire Zone rename, Sentry boundary breadcrumbs
- **iOS** ITMS-90683 fix, FFmpeg disabled in Xcode Cloud
- **Perf** lazy-loaded per-game sound/image registries, compressed PNG icons

## Conflict resolutions

- Cascade: deleted v1 `engine.ts`, `engine.shared.ts`, old e2e specs/tests (replaced by v2)
- Yacht i18n: 13 locale files — kept dev's `gameOver.scorecard` key and all VS-mode strings
- StarSwarm: kept Free Fire Zone rename, Sentry breadcrumbs, `useBackgroundMusic` registry refactor
- Binary assets: kept dev's compressed PNGs; expo `~55.0.25` → `~55.0.26`

## Root cause note

main was not merged back into dev after PR #1745 (May 19). Going forward: after each dev→main merge, run `git checkout dev && git merge main && git push`.

## Test plan

- [ ] gate-main-source passes (branch is `release/dev-to-main`)
- [ ] Maestro smoke CI passes on Android & iOS
- [ ] Playwright E2E suite passes
- [ ] Backend tests pass


wcmchenry3-stack:
## main/dev sync — lessons learned

**Root cause of the 40 conflicts:** After each previous dev→main merge, main's merge commit was never backported to dev. Git saw the histories as diverged, so the next release required a full three-way merge instead of a fast-forward.

**The fix (one step after every release):**
```bash
git checkout dev && git merge main && git push
```

This doesn't pull new features from main into dev — it just tells dev "I know about that merge commit." Without it, the divergence accumulates and conflicts multiply.

**Why this is safe:** Dev is always strictly ahead of main content-wise. The only thing on main that dev is missing is the merge commit itself, so this is almost always a no-op fast-forward with no conflicts.

**Why we used a `release/*` branch this time instead of resolving on dev:** Dev is the source of truth and we didn't want to risk touching it during conflict resolution. The release branch approach is the right pattern for that constraint — but it comes with overhead (extra branch, extra CI surface, and as we saw here, git's auto-merge can silently corrupt files in ways that only surface in CI).

**Going forward:** If the one-step sync above is run after each release, the next dev→main PR will be a clean fast-forward — no release branch, no conflicts, no surprises.

PR #1974, opened 2026-06-06
fix(yacht): AI hold/score bugs + joker bonus in VS scorecard

### LINKED ISSUE #1973: yacht: AI Joker decision rules — always apply to highest-value open category
## Problem

When the computer player rolls a Joker (all-same dice with Yacht already scored), its scoring decision should maximise value by picking the highest-value open category. The current `scoreMedium`/`scoreHard` functions use heuristics designed for normal rolls that fire before `bestInLegal`, causing suboptimal joker placement in edge cases.

**Example of bad behaviour (fixed in branch `claude/yacht-computer-player-logic-qIdd4`):**
- Player has `large_straight` already scored, rolls a Joker (e.g. five 6s).
- `jokerPossibleScores` returns all open lower categories with joker-inflated values: `full_house=25`, `small_straight=30`, `three_of_a_kind=30`, `four_of_a_kind=30`, `chance=30`.
- `scoreMedium` hard-codes "always take full_house" before checking the others — so it picks 25 pts instead of 30.

## Resolution

The fix in the linked branch adds an `isJokerRoll()` guard at the top of both `scoreMedium` and `scoreHard`. When a Joker is detected, the function immediately returns `bestInLegal(legal)`, which picks the highest value from the priority-filtered legal set already produced by `jokerPossibleScores`. The joker priority rules (mandatory upper → open lower → fallback upper) are enforced upstream in the engine, so `bestInLegal` is always choosing from the correct set.

## Priority order the AI now follows for a Joker

1. **Corresponding upper category open** → score there (engine forces this; AI sees only that option).
2. **Corresponding upper filled, open lower categories available** → pick the highest value among: `large_straight` (40) > `small_straight`/`three_of_a_kind`/`four_of_a_kind`/`chance` (30 each, face-dependent) > `full_house` (25).
3. **All lower filled** → any open upper (pick highest value via `bestInLegal`).

## Related changes in same branch

- `holdMedium` now guards Full House, 5-run, 4-run, and 3-run holds against already-scored categories (separate bug; AI was locking dice for patterns that were no longer useful).
- `VsScorecard` now shows a **Joker Bonus** row during gameplay, matching the existing solo `Scorecard` behaviour.

## Files changed

- `frontend/src/game/yacht/ai.ts`
- `frontend/src/components/yacht/VsScorecard.tsx`
- `frontend/src/screens/GameScreen.tsx`

wcmchenry3-stack:
$(cat <<'EOF'
## Summary

- **AI hold bugs fixed:** `holdMedium` no longer locks Full House dice when `full_house` is already scored (falls through to trips hold instead), and no longer pursues straight holds (5-run, 4-run, 3-run) when both `large_straight` and `small_straight` are already scored.
- **AI Joker scoring fixed:** `scoreMedium` and `scoreHard` now short-circuit to `bestInLegal` when a Joker is active, ensuring the AI always picks the highest-value legal category (e.g. `small_straight` at 30 pts rather than `full_house` at 25).
- **Joker bonus row in VS scorecard:** `VsScorecard` now shows a Yacht Bonus row during gameplay for both player and CPU, matching the existing solo `Scorecard` behaviour.

## Test plan

- [ ] Play VS mode and score Yacht, then roll a Joker — verify AI applies it to the highest-value open lower category (large_straight > small_straight/three_of_a_kind/chance > full_house)
- [ ] Fill `full_house` early in a VS game; roll a 3+2 — verify AI holds only the 3-of-a-kind, not all 5 dice
- [ ] Fill both straights; roll a 4-run — verify AI does not lock the run, re-rolls instead
- [ ] Score a Joker in VS mode — verify a "Yacht Bonus" row appears in the scorecard with the correct value for both player and CPU
- [ ] Solo mode scorecard joker bonus row still shows correctly (no regression)

Closes #1973

https://claude.ai/code/session_015AYXr7Hgu4jLLtoUWzw6ep
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_015AYXr7Hgu4jLLtoUWzw6ep)_

### DIFF
--- a/frontend/src/components/yacht/VsScorecard.tsx
+++ b/frontend/src/components/yacht/VsScorecard.tsx
@@ -15,9 +15,11 @@ export interface VsScorecardProps {
   playerRollsUsed: number;
   playerGameOver: boolean;
   playerUpperBonus: number;
+  playerYachtBonusTotal: number;
   playerTotalScore: number;
   cpuScores: Record<string, number | null>;
   cpuUpperBonus: number;
+  cpuYachtBonusTotal: number;
   cpuTotalScore: number;
   isAiTurn: boolean;
   onScore: (category: string) => void;
@@ -29,9 +31,11 @@ export default function VsScorecard({
   playerRollsUsed,
   playerGameOver,
   playerUpperBonus,
+  playerYachtBonusTotal,
   playerTotalScore,
   cpuScores,
   cpuUpperBonus,
+  cpuYachtBonusTotal,
   cpuTotalScore,
   isAiTurn,
   onScore,
@@ -248,6 +252,41 @@ export default function VsScorecard({
     );
   }
 
+  function renderJokerBonusRow() {
+    if (playerYachtBonusTotal === 0 && cpuYachtBonusTotal === 0) return null;
+    return (
+      <View style={[styles.vsRow, styles.vsBonusRow, { borderBottomColor: colors.border }]}>
+        <View style={styles.vsRowLeft}>
+          <Text style={[styles.vsBonusLbl, { color: colors.textMuted }]}>
+            {t("bonus.yachtLabel")}
+          </Text>
+        </View>
+        <View style={[styles.vsCell, { backgroundColor: "transparent" }]}>
+          <Text
+            style={{
+              color: playerYachtBonusTotal > 0 ? colors.bonus : colors.textMuted,
+              fontSize: 15,
+              fontWeight: "700",
+            }}
+          >
+            {playerYachtBonusTotal > 0 ? `+${playerYachtBonusTotal}` : "—"}
+          </Text>
+        </View>
+        <View style={[styles.vsCell, { backgroundColor: "transparent" }]}>
+          <Text
+            style={{
+              color: cpuYachtBonusTotal > 0 ? colors.bonus : colors.textMuted,
+              fontSize: 15,
+              fontWeight: "700",
+            }}
+          >
+            {cpuYachtBonusTotal > 0 ? `+${cpuYachtBonusTotal}` : "—"}
+          </Text>
+        </View>
+      </View>
+    );
+  }
+
   return (
     <ScrollView
       focusable={Platform.OS === "web"}
@@ -276,6 +315,7 @@ export default function VsScorecard({
         {(LOWER_CATEGORY_KEYS as readonly string[]).map((key, i) =>
           renderCategoryRow(key, "lower", i === LOWER_CATEGORY_KEYS.length - 1)
         )}
+        {renderJokerBonusRow()}
         {renderSubtotalRow(t("vsMode.lowerSubtotal"), playerLowerSubtotal, cpuLowerSubtotal, false)}
       </View>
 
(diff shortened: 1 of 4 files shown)


wcmchenry3-stack:
## Code Review

### Overview

This PR fixes three distinct bugs in Yacht VS mode:
1. AI hold logic was locking dice for already-scored categories (Full House, straights)
2. AI scoring logic ignored Joker priority rules, picking suboptimal categories
3. VS scorecard was missing the Yacht Bonus row that solo mode already had

---

### Code Quality

**`ai.ts` — Hold fixes (solid)**

- The `isJokerRoll` helper is clean and correctly guards on `scores.yacht === 50` with `c.size === 1`. The `(dice[0] ?? 0) > 0` guard prevents false positives on empty arrays.
- The `straightOpen` variable is well-named and avoids repeating the OR expression across three guards.
- The full-house fix (`isOpen(scores, "full_house")`) is correct — falls through to the Trips branch below, which is the right behavior.

**`ai.ts` — Score fixes (one concern)**

The Joker short-circuit is placed *before* the Yacht check in both `scoreMedium` and `scoreHard`:

```ts
if (isJokerRoll(dice, scores)) return bestInLegal(legal) ?? open[0]!;
// Yacht — always take 50 pts
if ("yacht" in legal && ...) return "yacht";
```

- `isJokerRoll` requires `scores.yacht === 50`, so when a Joker fires, yacht is already scored — the Yacht branch would never be reached anyway. The ordering is safe, but the co-location with the comment `// Always take Yacht` could confuse a future reader into thinking these can conflict. A brief inline note (e.g. `// yacht already scored when isJokerRoll is true`) would prevent that misreading — or swap the ordering so the Joker check comes *after* the unreachable Yacht check for clarity.

**`VsScorecard.tsx` — Joker bonus row**

- Conditional `if (playerYachtBonusTotal === 0 && cpuYachtBonusTotal === 0) return null` correctly hides the row in all non-Joker games.
- Inline styles (`fontSize: 15`, `fontWeight: "700"`) are used here, whereas the file likely uses a StyleSheet elsewhere. Minor inconsistency, but worth checking if the existing bonus/subtotal rows use the same pattern.
- The `colors.bonus` token is used — confirm this key exists in the theme object (not visible in the diff). If it's missing, it silently renders `undefined` color on some platforms.

**`GameScreen.tsx`**

- Straightforward prop threading — clean.

---

### Potential Issues

1. **`open[0]!` non-null assertion** — `bestInLegal(legal) ?? open[0]!` — if `legal` is empty *and* `open` is also empty (theoretically possible at game-end edge), this throws at runtime. Low risk given game-end state shouldn't trigger `scoreAI`, but worth a note.

2. **No `holdHard` changes** — `holdEasy` and `holdMedium` are fixed, but `holdHard` is not shown in the diff. If `holdHard` has equivalent straight/full-house lock bugs, they'd remain. Worth a quick check.

3. **Test coverage** — the test plan is manual. Given these are pure functions (`holdMedium`, `scoreMedium`, `scoreHard`), unit tests for the exact regression cases (fill `full_house`, roll 3+2; fill both straights, roll 4-run; Joker with various open categories) would prevent future regressions here. Not a blocker, but this logic is subtle enough to warrant it.

---

### Summary

The fixes are correct and well-scoped. The main things worth addressing before merge:

- **Verify `holdHard` doesn't have the same straight/full-house bugs** (not touched in this PR).
- **Confirm `colors.bonus` token exists** in the theme.
- The `open[0]!` assertion is low-risk but worth awareness.
- Consider unit tests for the AI regression cases.

Issue #2033, opened 2026-06-12
feat(solitaire): hint engine (getHintMoves/isProductiveMove/applyHint)


wcmchenry3-stack:
Part of epic #2023 (story B0).

## Scope

Port FreeCell's hint pattern (`frontend/src/game/freecell/engine.ts`: `getHintMoves`, `isProductiveMove`, `applyHint`) into `frontend/src/game/solitaire/engine.ts`, adapted to Klondike's move types. Priority order: foundation moves → tableau-to-tableau moves that reveal a face-down card → waste-to-tableau → other legal tableau moves. Filter non-productive reversible swaps per the FreeCell precedent (#1295). Engine-only — the UI button lands in B3. Engine stays pure (no React/IO).

## Acceptance criteria

- [ ] `getHintMoves` returns foundation move first when one exists
- [ ] Returns the face-down-revealing tableau move when no foundation move exists
- [ ] Returns waste-to-tableau before non-revealing tableau shuffles
- [ ] Reversible non-productive swaps excluded
- [ ] No-moves case returns empty list / `applyHint` leaves state unchanged
- [ ] `applyHint` sets `state.hint` without mutating the input state (immutability preserved)

## Test plan

- **Unit (Jest, `solitaire/__tests__/engine.test.ts`):** one scripted board per priority tier; productive-move filter cases; no-moves case; immutability check.
- **Playwright:** existing `solitaire-*.spec.ts` green (engine addition only, no UI yet). Full hint-flow spec lands with B3.
- **Maestro:** existing `flows/solitaire/smoke.yaml` green (no UI change). Device hint flow lands with B3.

## Dependencies

None — unblocks B3 (hint button).

**Size:** M | **Implementation model:** `sonnet` (priority ordering + productive-move judgment for Klondike's hidden-information layout)

Issue #2023, opened 2026-06-12
Epic: Solitaire & FreeCell experience — hybrid tap interaction, hints, FreeCell parity


wcmchenry3-stack:
## Goal

Bring Solitaire and FreeCell up to the mobile card-game gold standard — the **hybrid interaction model** — and close FreeCell's feature gaps vs Solitaire.

**Hybrid interaction model:**
1. **Single-tap smart auto-move** (the workhorse, ~80% of moves): tapping a card fires a deterministic priority ladder — foundation → tableau (preferring moves that reveal face-down cards / build runs) → empty column → free cell last (FreeCell only).
2. **Two-tap targeted choice** (ambiguity resolver): when multiple equal-priority destinations exist, keep the select+highlight flow — first tap selects and ghosts valid destinations, second tap chooses.
3. **Drag-and-drop** (tactile override): stays, with a generous ~35% snap radius on drop targets and elastic snap-back.

**FreeCell micro-UX:** supermove limit pre-checked at interaction start (shake + toast "Not enough free spaces to move this stack"); double-tap sends to first open free cell; tap targets weighted to the bottom exposed edge of overlapped cards.

**Feature parity:**
- Solitaire gains a **hint system** (port FreeCell's `getHintMoves` / `isProductiveMove` / `applyHint` pattern) with a −20 pt penalty.
- FreeCell gains **per-device stats**, a **leaderboard client** (backend `POST /freecell/score` + `GET /freecell/leaderboard` already exist — frontend-only), and a **winnable-seed bank** (Python DFS solver mirroring `backend/scripts/gen_solitaire_seeds.py`) so players never get an unsolvable deal.

## Stories

**Solitaire hints** (`feat/solitaire-hints`)
- [x] #2033 — B0: hint engine in solitaire/engine.ts (M, **sonnet**)
- [ ] #2036 — B3: hint button + −20 score penalty (S, **haiku**, blocked by #2033)

**Smart tap interaction** (`feat/smart-tap`)
- [ ] #2034 — B1: generous snap radius in shared DragContext (S, **haiku**)
- [ ] #2037 — B4: FreeCell smart single-tap auto-move + supermove pre-check (M, **sonnet**)
- [ ] #2039 — B6: Solitaire smart single-tap auto-move (M, **sonnet**, after #2037 to reuse its helper)

**FreeCell parity** (`feat/freecell-parity`)
[...]

PR #2046, opened 2026-06-12
feat(solitaire): hint button with −20 score penalty

### LINKED ISSUE #2036: feat(solitaire): hint button + score penalty
Part of epic #2023 (story B3).

## Scope

Wire the B0 hint engine (#2033) into `frontend/src/screens/SolitaireScreen.tsx`: add a "Hint" button to the header row using FreeCell's existing hint-button visual style, render hint highlighting on the source and destination (same pattern as `FreeCellBoard`'s hint rendering), and apply a **−20 point** penalty per hint used (named constant next to the other scoring constants in the engine; score still floors at 0). Button disabled when the game is won or auto-complete is running.

## Acceptance criteria

- [ ] Hint button highlights the top-priority move's source + destination
- [ ] Score decreases by 20 per hint (floored at 0); move counter unaffected
- [ ] Disabled during auto-complete and after win
- [ ] Accessibility: button labeled, hint announced via live region (match FreeCell's a11y pattern)

## Test plan

- **Unit (Jest):** screen test — button dispatches `applyHint`, penalty applied, disabled states; engine test for the penalty constant.
- **Playwright:** new `e2e/tests/solitaire-hint.spec.ts` — tap Hint, assert highlight appears and score drops by 20; assert disabled state during auto-complete; a11y attributes present.
- **Maestro:** new `e2e/maestro/flows/solitaire/hint.yaml` — launch, open solitaire, tap Hint, assert highlight renders on device.

## Dependencies

Blocked by #2033 (B0).

**Size:** S | **Implementation model:** `haiku` (UI wiring on an existing pattern)

wcmchenry3-stack:
Closes #2036 (story B3 of epic #2023).

Wires the hint engine (#2033) into SolitaireScreen: header Hint button (FreeCell's visual + a11y pattern), golden highlight on the hint move's source/destination via new props on TableauPile/FoundationPile/StockWastePile, hint cleared on any real move. `HINT_PENALTY = 20` lives engine-side in `applyHint` (floored at 0) so the rule is unit-testable. `action.hint` key added to all 13 locales. Disabled during auto-complete and after win.

**Tests:** engine penalty/floor/clear-on-move + screen disabled-state units (96 passing); new `e2e/tests/solitaire-hint.spec.ts` (highlight, −20, floor-at-0, disabled, clear-on-move — selectors verified against the implementation); new Maestro `solitaire/hint.yaml`.

### DIFF
--- a/e2e/maestro/flows/solitaire/hint.yaml
+++ b/e2e/maestro/flows/solitaire/hint.yaml
@@ -1,0 +1,14 @@
+appId: com.buffingchi.games
+---
+- runFlow: ../_shared/launch.yaml
+- runFlow:
+    file: ../_shared/navigate-to.yaml
+    env:
+      gameSlug: "solitaire"
+      screenLabel: "Solitaire"
+- assertVisible:
+    id: "solitaire-hint-button"
+    timeout: 8000
+- tapOn:
+    id: "solitaire-hint-button"
+- takeScreenshot: solitaire-hint
--- a/e2e/tests/solitaire-hint.spec.ts
+++ b/e2e/tests/solitaire-hint.spec.ts
@@ -28,13 +28,13 @@ function stockCards(excluded: string[]) {
 
 test.describe("Solitaire — hint button", () => {
   // Board:
-  //   col 0: [3♥(fu)]  — source card for hint (can move to foundation)
-  //   col 1: [A♠(fu)]  — aces start foundations
-  //   col 2: [2♠(fu)]  — in place for tableau-to-foundation move
+  //   col 0: [3♥(fu)]  — occupied column (cannot move to foundation; A♥/2♥ not yet there)
+  //   col 1: [A♠(fu)]  — first hint source: A♠ → spades foundation
+  //   col 2: [2♠(fu)]  — in place for tableau-to-foundation once A♠ is placed
   //   col 3: [K♣(fu)]  — destination for possible tableau moves
   //   col 4: [Q♦(fu)]  — another available card
-  //   col 5: [2♥(fu)]  — will move to hearts foundation
-  //   col 6: [2♦(fu)]  — will move to diamonds foundation
+  //   col 5: [2♥(fu)]  — will move to hearts foundation after A♥ is placed
+  //   col 6: [2♦(fu)]  — will move to diamonds foundation after A♦ is placed
   const BOARD_STATE_WITH_HINT = {
     _v: 1,
     drawMode: 1,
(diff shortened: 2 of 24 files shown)


wcmchenry3-stack:
## Code Review — feat(solitaire): hint button with −20 score penalty

### Overview

Wires the existing hint engine into `SolitaireScreen`: adds a Hint button to the header alongside Undo, applies a `HINT_PENALTY = 20` in `applyHint()` (floored at 0), highlights source/destination cards via new props on the three pile components, clears hint state on any real move, and ships i18n for all 13 locales. Engine-unit tests + screen-unit tests + a Playwright E2E spec + a Maestro smoke flow are included.

---

### Correctness Issues

**1. Hint button stays visually enabled when no hints exist**

`hintDisabled` (`SolitaireScreen.tsx:722`) only gates on `isComplete` and `autoCompleting`. But `handleHint` has an additional early-return guard:
```ts
if (getHintMoves(state).length === 0) return;
```
So when no moves are available, the button *looks* clickable but silently does nothing. `getHintMoves` should be memoized and folded into `hintDisabled` so the visual state matches the functional state.

**2. `foundation-to-tableau` source has no highlight**

In `hintSourceIndex` (~line 735), the `foundation-to-tableau` case returns `undefined`, so the source foundation pile receives no highlight. The destination tableau column gets highlighted, but the user has no visual cue *which* foundation card to move. `FoundationPile` already accepts `hintDestination` — a `hintSourceSuit` prop (or equivalent) would close this gap.

**3. E2E board state comment is wrong**

`solitaire-hint.spec.ts:57`:
```
//   col 0: [3♥(fu)]  — source card for hint (can move to foundation)
```
With all foundations empty, 3♥ cannot go to the hearts foundation — A♥ and 2♥ would need to be there first. The actual first hint from this board is A♠ → spades foundation. Comment-only bug, but will mislead the next person debugging a failing E2E.

---

### Code Quality

**4. `getHintMoves` is called twice per tap**

`handleHint` calls `getHintMoves(state)` as a guard, then `applyHint` calls it again internally. Either inline the check or memoize `getHintMoves` to avoid running move-finding twice on every tap.

**5. Confusing variable naming**

`hintSourceIndex` stores a *column* index; `hintSourceCardIndex` stores a *card* index within that column. The names read as two "card indexes." Renaming to `hintSourceCol` / `hintSourceCardIndex` would make the prop mapping at `TableauPile` self-evident.

**6. E2E highlight test doesn't actually verify the highlight**

The primary visual feature — the 3px bonus-colored border — is explicitly skipped in the test with a comment saying it can't be checked. Playwright can query computed styles, but more practically, adding a `testID` to the highlighted card (e.g. `solitaire-hint-source`) would let the test assert `toBeVisible()` on the element rather than just the board container.

---

### Minor Notes

- **Lint:** run `cd frontend && npx eslint src/screens/SolitaireScreen.tsx src/game/solitaire/components/` before merging — the inline style object on `StockWastePile.tsx:409` (`style={hintSource ? { borderColor: colors.bonus, ... } : undefined}`) will likely trigger the `react-native/no-inline-styles` rule that the others correctly avoid by hoisting to a `const hintStyle`.
- **Maestro flow** is a pure smoke test (tap + screenshot, no score assertion). Fine as a gate but worth noting given the Playwright spec also concedes the highlight gap.
- **No accessibility announcement for the −20 penalty** — screen reader users won't hear that points were deducted. An `accessibilityLiveRegion` on the score display would address this.
- **`handleHint` guard duplicates `hintDisabled`** — the `state === null || state.isComplete || autoCompleting` checks exist in both; they'll drift if one is updated without the other.

---

### What's Working Well

- `HINT_PENALTY` is engine-side and unit-tested — correct separation of concerns.
- `hint: undefined` injected into `finalizeAfterMove` cleanly clears hint on any real move without screen-level bookkeeping.
- `hintDestination` / `hintSource` / `hintIndex` props are additive and backwards-compatible.
- All 13 locales updated atomically with the feature.
- `hintStyle` defined once per component, not recreated in JSX — right pattern for RN.

---

### Summary

Two real correctness gaps worth addressing before merge: the button not visually reflecting "no hints available," and the missing source highlight for `foundation-to-tableau` moves. Fix the board comment in the E2E fixture. The inline style on `StockWastePile` needs to be hoisted to pass lint. The double `getHintMoves` call and variable naming are cleanup-level.

Issue #2128, opened 2026-06-19
revert(solitaire,freecell): remove smart auto-move — restore tap-to-select interaction model


wcmchenry3-stack:
## Problem

PRs #2037 and #2039 added a "smart single-tap auto-move" feature to FreeCell and Solitaire respectively. When a card has one unambiguous valid destination, a single tap immediately executes the move without the user selecting a destination first.

This was shipped without product sign-off and **breaks the intended interaction model**. Cards now "jump" on first tap, which is unexpected and feels like a bug.

## Expected behavior (original, pre-#2037/#2039)

1. User taps a card → card becomes **selected** (highlighted)
2. User taps a valid destination → card **moves**
3. User taps the same card again → card **deselects**
4. User taps an invalid destination → **illegal move** feedback (shake)

This model should be the primary interaction. Drag-and-drop is a secondary interaction (tracked in epic #2127).

## What to revert

**FreeCell** — `FreeCellBoard.tsx`: remove `resolveAutoMove` call in `handleTableauCardPress` and `handleFreeCellPress`. Restore the plain `selection`-based state machine.

**Solitaire** — `SolitaireScreen.tsx`: remove `resolveAutoMove` call and the single-tap path in `handleTableauCardPress`. Restore the prior select-then-confirm flow.

Both `resolveAutoMove` functions in the respective engines can be left in place (they're used by the hint system) — only the UI call sites need to be removed.

## Acceptance criteria

- [ ] Tapping a face-up card selects it (visual highlight appears)
- [ ] Tapping a legal destination moves the card
- [ ] Tapping the selected card again deselects it
- [ ] Tapping an illegal destination triggers shake feedback, selection is cleared
- [ ] Existing Playwright tap-move tests still pass (they test the selection flow, not auto-move)
- [ ] Double-tap to foundation still works (this is a separate, desirable shortcut)

## Epic

#2127

wcmchenry3-stack:
Resolved by #2133 (fix(drag): tap-to-select + Fabric container measurement), merged 2026-06-19.

Issue #2127, opened 2026-06-19
epic(drag): reliable card drag-and-drop for Solitaire + FreeCell on iOS/Android


wcmchenry3-stack:
## Epic Overview

Card drag-and-drop has **never worked reliably** in Klondike Solitaire or FreeCell on iOS/Android. This is not a regression from a single bad commit — iOS broke on day one, every subsequent "fix" PR iterated on a broken base, and problems compounded. This epic tracks a ground-up repair with iOS/Android as the primary constraint.

Separately, the tap interaction was broken by smart auto-move (#2037, #2039) — tracked in #2128.

---

### Why iOS never worked from the start

The original April 25 implementation (`e454962f`) had the correct architecture:
- Source card `opacity: 0` while dragging (hidden, not dimmed)
- Plain `View` ref with `.measure()` for container position
- `GestureDetector` wrapping each card

But it had an undetected iOS flaw: **`View.measure()` returns stale/zero values on iOS on the first layout pass** before the native view tree settles. Container offset stayed at (0,0), so the ghost always appeared at the top-left of the `DragContainer`. On web and Android this was fine; on iOS it was broken immediately.

### Why April 29 made things worse

Four days after launch, `6b3f2716` addressed three real iOS problems:

1. `View.measure()` stale on first layout → switched to `rnMeasure()` (Reanimated worklet) at drag-start
2. RNGH 2.30.0 regression prevented pan from exceeding `minDistance` on iOS → switched `Gesture.Simultaneous` to `Gesture.Race`
3. Container-offset race (DragContainer.onLayout hadn't fired before first drag) → switched container to `Animated.View` + `AnimatedRef`

These were real problems. But the fix introduced a new one: **`AnimatedRef.current` is a Reanimated shadow node on Fabric (new arch), not a real view.** The `DragContainer.onLayout` handler calls `containerRef.current?.measureInWindow(...)` on it, which silently no-ops. `containerOffsetX/Y` never gets set.

The worklet path (`rnMeasure(containerRef)` in `DraggableCard.onStart`) is the correct approach, but can return null before the first native layout pass, leaving the ghost at the wrong offset on the first drag.

### Why subsequent refactors didn't fix it

Every PR after April 29 targeted symptoms without touching the ghost position or measurement reliability. Two more regressions were added along the way:

[...]

Issue #2037, opened 2026-06-12
feat(freecell): smart single-tap auto-move + supermove pre-check


wcmchenry3-stack:
Part of epic #2023 (story B4).

## Scope

Implement the hybrid interaction model's workhorse in `frontend/src/components/freecell/FreeCellBoard.tsx`. When no selection is active and the player taps a movable card/run, fire a smart auto-move down a deterministic priority ladder:

1. **Foundation** (legal next card of suit) — move immediately
2. **Tableau** — legal column moves; prefer the destination that builds the longest run
3. **Empty tableau column**
4. **Free cell** — last resort only

If exactly one destination exists at the winning priority tier → execute immediately. If multiple equal-priority destinations exist → fall back to the current two-tap flow: keep the selection and highlight/ghost all valid destinations; second tap chooses. Drag-and-drop is untouched as the manual override.

Micro-UX from the research doc:
- **Supermove pre-check at interaction start:** if the tapped run exceeds `(1 + freeCells) × 2^emptyColumns`, do not start the interaction — shake the stack + toast *"Not enough free spaces to move this stack"* (engine already computes the limit; today it's only checked at move validation).
- **Double-tap = send to first open free cell** (already partially present — confirm and keep).
- **Tap targets weighted to the bottom exposed edge** of overlapped cards so the visible sliver selects the right card.

## Acceptance criteria

- [ ] Single tap with one best destination moves the card with no second tap
- [ ] Ambiguous destinations → selection + destination highlighting, second tap resolves
- [ ] Oversized run tap → shake + toast, no selection state entered
- [ ] Double-tap → first open free cell; drag unchanged
- [ ] Existing sounds/haptics fire on auto-moves (cardPlace / supermove)

## Test plan

- **Unit (Jest, `components/freecell/__tests__/`):** priority-ladder resolution on scripted boards (foundation beats tableau, free cell last); ambiguity fallback; supermove pre-check gating.
- **Playwright:** update `freecell-card-move.spec.ts` for the auto-resolve path (single tap moves); add ambiguous-board case proving two-tap fallback; new supermove-rejection toast assertion; `freecell-errors.spec.ts` re-verified.
- **Maestro:** new `e2e/maestro/flows/freecell/tap-automove.yaml` — single-tap an ace, assert it lands on a foundation on device; attempt an oversized run, assert the toast renders (native toast + gesture surface).

## Dependencies

[...]

PR #2133, opened 2026-06-19
fix(drag): tap-to-select + Fabric container measurement (#2126/#2128)

### LINKED ISSUE #2126: fix(drag): repair solitaire + freecell drag — opacity, container measurement, hitSlop, testID
## Context

See epic #2127 for full history. This issue covers the four concrete code changes needed to make drag work, plus the Maestro assertion upgrade that gates them.

---

## Implementation Order

**Step 1 first — Maestro assertions before any code fix.** Every previous drag regression slipped through CI because the Maestro flows had wrong testIDs (element never found, swipe silently skipped) and only checked that the board was still visible after a swipe — not that the card actually moved. Fix the tests before fixing the code, so CI is a real gate, not theatre.

---

### FreeCell (`e2e/maestro/flows/freecell/drag.yaml`)

Problem: `freecell-col-0-top` and `freecell-col-1-top` don't exist. The swipe targets are never found, Maestro skips them silently, and the test passes regardless of whether drag works.

Fix requires two parts:

**a) Add `-top` testID to `TableauColumn.tsx`:**
```tsx
// frontend/src/components/freecell/TableauColumn.tsx
testID={isTop
  ? `freecell-col-${colIndex}-top`
  : `freecell-col-${colIndex}-card-${cardIndex}`}
```

**b) Update Maestro YAML to assert the card actually moved:**
```yaml
# After swipe from freecell-col-0-top to freecell-slot-0:
- assertVisible:
    id: "freecell-slot-0-card"   # testID on the card now in the slot
    timeout: 2000
# NOT just: assertVisible freecell-slot-0 (the empty slot is always visible)
```

This requires `FreeCellSlot` to expose a `testID` on the card when occupied (e.g. `freecell-slot-${cellIndex}-card`).

### Solitaire (`e2e/maestro/flows/solitaire/drag.yaml`)

The testIDs are correct (`solitaire-waste-top`, `solitaire-tableau-0`), but the post-swipe assertion only checks a screenshot — it doesn't verify the card moved.

Fix: after swiping waste-top → tableau-0, assert that tableau column 0 has a card (e.g. assert the waste pile top changed, or that the tableau column label updated to "1 card").

---

## Step 2 — Source card opacity: 0 during drag (was 0.6)

**File:** `frontend/src/game/_shared/drag/DraggableCard.tsx`

```tsx
// Current (wrong — user sees two cards: dim source + full ghost)
const dimmedStyle = useAnimatedStyle(() => ({
  opacity: beingDragged ? 0.6 : 1,
}));

// Fix — source card invisible; ghost overlay renders it at full opacity
const dimmedStyle = useAnimatedStyle(() => ({
  opacity: beingDragged ? 0 : 1,
}));
```

---

## Step 3 — Container offset measurement via `runOnUI` + `measure`

**File:** `frontend/src/game/_shared/drag/DragContainer.tsx`

Current `onLayout` calls `containerRef.current?.measureInWindow(...)` which silently no-ops on Fabric because `containerRef.current` is a Reanimated shadow node, not a real view.

`runOnUI` + `measure` is the only path that works on all four combinations: iOS Paper, iOS Fabric, Android Paper, Android Fabric.

```tsx
import { runOnUI, measure } from 'react-native-reanimated';

const onLayout = useCallback(
  (e: LayoutChangeEvent) => {
    runOnUI(() => {
      'worklet';
      const m = measure(containerRef);
      if (m) {
        containerOffsetX.value = m.pageX;
        containerOffsetY.value = m.pageY;
      }
    })();
    externalOnLayout?.(e);
  },
  [containerRef, containerOffsetX, containerOffsetY, externalOnLayout]
);
```

---

## Step 4 — `hitSlop` on the gesture, not the wrapper View

**File:** `frontend/src/game/_shared/drag/DraggableCard.tsx`

`hitSlop` on `Animated.View` expands the native touch area but does **not** expand the RNGH `GestureDetector`'s recognition area. Buried card stripes can't be dragged from their edges.

```tsx
const pan = Gesture.Pan()
  .minPointers(1)
  .minDistance(5)
  .enabled(draggable)
  .hitSlop(hitSlop ?? {})
  ...

const tap = Gesture.Tap()
  .maxDistance(8)
  .hitSlop(hitSlop ?? {})
  ...

// Remove hitSlop from the Animated.View wrapper
<Animated.View ref={viewRef} testID={testID} style={[style, webHitSlopStyle, dimmedStyle]}>
```

---

## Acceptance criteria

- [ ] `freecell-col-0-top` testID resolves to the top card of column 0
- [ ] Maestro freecell drag flow asserts card is in slot (not just board visible)
- [ ] Maestro solitaire drag flow asserts card count changed after drag
- [ ] Ghost card appears at exact position of dragged card on first drag attempt (iOS + Android)
- [ ] Source card slot is invisible (opacity 0) while dragging
- [ ] Drop on valid target commits the move — verified in Maestro CI
- [ ] Drop on invalid target snaps back — verified in Maestro CI
- [ ] No drag change merges without device verification OR Maestro passage

## Epic

#2127

wcmchenry3-stack:
## Summary

- **DragContainer**: Fixed container measurement via `runOnUI + Reanimated.measure` — the previous `measureInWindow` on `AnimatedRef.current` silently no-ops on Fabric/new arch (iOS & Android), causing the ghost card to appear at position (0,0)
- **DraggableCard**: Source card opacity changed from 0.6 → 0 while dragged (eliminates double-image); `hitSlop` moved from `Animated.View` to gesture `.hitSlop()` API (View hitSlop doesn't expand RNGH gesture area)
- **Tap-to-select restored**: Reverted `resolveAutoMove` smart single-tap from #2037/#2039 — single tap now selects, tap destination executes move, double-tap shortcut to foundation preserved
- **Maestro E2E**: Fixed FreeCell drag flow — was referencing non-existent testIDs, producing false-green CI; added `freecell-col-N-top` / `freecell-slot-N-card` testIDs and real move assertions

Closes #2126, #2128. Relates to epic #2127.

## Why this didn't work before (historical context)

iOS/Android drag broke at three layers simultaneously:
1. Measurement: `AnimatedRef.current.measureInWindow()` → silent no-op on Fabric
2. Visual: 0.6 opacity source card showed alongside ghost (double image)
3. UX: Auto-move on first tap meant no selection step, making moves feel random

## Test plan

- [ ] 106 unit tests pass (`FreeCellBoard`, `SolitaireScreen`, `TableauPile`, `DraggableCard`, `DragContainer`)
- [ ] `FreeCellBoard.smartTap.test.tsx` deleted (tests removed feature)
- [ ] Maestro FreeCell drag flow passes on device (requires physical iOS or Android)
- [ ] Maestro Solitaire drag flow passes on device
- [ ] Manual: drag card in Klondike — ghost appears at correct position, source card invisible
- [ ] Manual: drag card in FreeCell — same
- [ ] Manual: tap card → highlight; tap destination → move; double-tap top card → foundation
- [ ] No regression in Hearts, Yacht, Mahjong (no drag changes in those games)

https://claude.ai/code/session_01SLW1hn7kVEALJiwsPtwg5X


wcmchenry3-stack:
## Code Review

**14 files changed, +85/-399. Closes #2126, #2128.**

---

### What's Good

- **Fabric measurement fix is correct.** Replacing `requestAnimationFrame + measureInWindow` on `containerRef.current` with `runOnUI(() => { "worklet"; measure(containerRef) })()` is the right call. `measureInWindow` on an `AnimatedRef` is a known silent no-op on the new architecture. The comment in `DragContainer.tsx` explains the why clearly.
- **`hitSlop` moved to gesture API.** The old approach of setting `hitSlop` on the `Animated.View` wrapper doesn't expand the RNGH gesture hit area — that's a known RNGH behavior. Moving it to `.hitSlop()` on both the `Pan` and `Tap` gestures is correct.
- **Opacity 0 is the right choice.** When a ghost overlay renders the dragged card, 0.6 opacity on the source creates a double-image. Making it fully invisible is semantically correct.
- **Maestro fixes are meaningful.** Asserting `freecell-slot-0-card` (card is in the slot) rather than `freecell-slot-0` (slot exists) is the difference between a real test and a no-op.

---

### Issues & Suggestions

**1. `resolveAutoMove` may now be dead code in the engine — Medium**

The diff removes all call sites of `resolveAutoMove` from `FreeCellBoard.tsx` and `SolitaireScreen.tsx`. If there are no other callers, the function (and its exported type) are now dead code in `engine.ts`. Worth deleting or flagging for a follow-up, or the PR description should note intentional retention.

**2. `.hitSlop(hitSlop ?? {})` passes `{}` instead of skipping — Minor**

```ts
// DraggableCard.tsx
.hitSlop(hitSlop ?? {})
```

When `hitSlop` is `undefined`, this passes `{}` to RNGH. Passing an empty object is likely treated as zero-slop, which is functionally equivalent to no hitSlop, but it's subtly different from not calling `.hitSlop()` at all. Not a bug, but could be cleaner:

```ts
const pan = Gesture.Pan().minPointers(1).minDistance(5).enabled(draggable);
if (hitSlop) pan.hitSlop(hitSlop);
```

**3. Solitaire Maestro second drag may fail on an empty waste — Minor**

`solitaire/drag.yaml` adds a second swipe from `solitaire-waste-top` to `solitaire-tableau-1`. If the first drag succeeds and the deal happened to have only one waste card, the waste would be empty and `solitaire-waste-top` wouldn't exist for the second swipe. The deal is random, so this could cause intermittent CI failures. Consider guarding the second drag or ensuring the test state has ≥2 waste cards.

**4. Supermove rejection toast is silently removed — Observation (not a bug)**

`FreeCellScreen.tsx` removes `onSupermoveRejected={() => showToast(t("freecell:error.supermoveRejected"))}`. Users can now select a run that exceeds supermove capacity and only discover it's invalid when they tap a destination (where `tryMove`/`validateMove` rejects it and fires the illegal-move animation). Feedback is still there — just deferred and less explicit. Worth noting if the UX regression is intentional.

**5. `TableauPile` hitSlop test weakening — Minor (justified)**

The comment correctly explains that RNTL v14 only exposes host elements, so gesture-level hitSlop can't be unit-tested. The Maestro E2E note is the right call — just make sure the Maestro flow actually exercises the buried-card tap scenario or this gap stays uncovered.

---

### Summary

Approve with minor notes. The Fabric measurement fix is solid and the test/Maestro cleanup is good. Main things to confirm before merge: (1) check whether `resolveAutoMove` is now dead code in the engine and clean it up if so, and (2) guard the Maestro second drag against an empty waste pile.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2232, opened 2026-07-19
fix(deps): revert expo group bump that breaks Android native build

wcmchenry3-stack:
## Summary
- Reverts commit 01fd271d (#2181, expo group bump to SDK 57) to restore `dev` to a known-good state.
- After #2181 merged, `expo-modules-core`'s native Android code started requiring a JSI API (`facebook::jsi::ArrayBuffer::tryGetMutableBuffer`) that only exists in react-native 0.86+. `dev` is still pinned to react-native 0.85.3, so `android-release-smoke` now fails on any PR built against `dev` — confirmed independently via #2185 (an unrelated @sentry/react-native bump) hitting the identical native compile error.
- The react-native 0.86 bump that would fix this (#2183) is itself blocked: it breaks `react-native-gesture-handler@2.31.2` (upstream requires a 3.x rewrite — new hook-based API, not a safe drop-in), and trips `reactVersionMatch.test.ts`, a regression guard added after a past production incident (Sentry BC_GAMES-46).
- Reverting here to unblock the dependency queue; the SDK 57 upgrade should be redone later as one coordinated piece of work (Expo SDK + react-native + gesture-handler/reanimated/worklets together), not as a routine Dependabot bump.

## Verification
- [x] `npm run lint` — clean (1 pre-existing unrelated warning)
- [x] `npm run format:check` — clean
- [x] `npx jest --ci` — full suite passes (3021 passed; initial run had 6 timeout flakes that passed in isolation/rerun, confirmed unrelated to this change)
- [x] `npx expo install --check` — matches the exact pre-#2181 baseline mismatch set (all pre-existing/approved overrides)

## Test plan
- [ ] CI green, especially `android-build-check` / `android-release-smoke`
- [ ] Confirm no other merged PR depended on SDK 57

🤖 Generated with Claude Code


wcmchenry3-stack:
Note: `Podfile.lock freshness` fails here for the same reason it silently failed (non-blocking) on the original #2181 — no CocoaPods/macOS available in this environment to run `pod install`. Since Podfile.lock wasn't touched by #2181 either, this revert is consistent with that state. Flagging in case someone with Mac tooling wants to regenerate it properly in a follow-up.

Issue #2238, opened 2026-07-19
feat(hearts-sim): sim gate v2 — paired-deal replay, SPRT gates, conditional metrics, CI wiring


wcmchenry3-stack:
Part of: #2233

## In plain English
Before we can safely tune any Hearts AI behavior (this epic proposes several such changes), we need a simulation gate we can actually trust and that runs automatically. Today's harness is manual-only, uses fixed sample-size thresholds that produce noisy pass/fail results, and mixes up "did this event happen" with "how often, out of how many chances." This issue redesigns the gate's methodology; it does not fix the specific broken metrics already tracked in #2204 — this builds the next-generation harness on top of that fix.

## User story
As a developer changing Hearts AI behavior, I want an automated, statistically sound simulation gate in CI so that a strategy change either provably doesn't regress persona behavior or is caught before merge, instead of relying on a human reading console output from a manual script run.

## Context
`scripts/simulate-hearts.ts` (550 lines) currently:
- Is **not wired into CI** — confirmed by inspection: no `.github/workflows/*.yml` references `simulate-hearts`, and no `npm`/`package.json` script invokes it as part of a checked pipeline. It's a manual developer tool only.
- Uses fixed-N batches (6 persona-matchup batches × 3,000 seeded games each, per `docs/hearts-qa-report-2026-07.md`) with hardcoded threshold checks (`zTest`/`sigLabel` at `scripts/simulate-hearts.ts:295-316`, threshold assertions around line 442+) — a single run's pass/fail is subject to sampling noise at the borderline (HRT-3 in the QA report: the "Cautious vs Schemer" built-in check was found to have its *direction* backwards, only masked because the tolerance happened to pass by chance on a borderline effect).
- Deals are already seeded and deterministic (`simulateGame(difficulties, seed)` calls `setRng(createSeededRng(seed))` at `scripts/simulate-hearts.ts:54`), which makes duplicate-deal replay straightforward to add — no engine changes needed, just orchestration around existing seeding.
- Mixes unconditional and conditional metrics without denominators — e.g. HRT-1 in the QA report found the "earlyMoon Success" metric (`simulate-hearts.ts:452-497`) divides *all* completed moon shots by a narrow-trigger denominator instead of pairing each activation with its own outcome, inflating the reported rate from a true ~4.9% to 48-80%.

[...]

Issue #2234, opened 2026-07-19
feat(hearts-ai): rank-aware moon-attempt hand evaluation


wcmchenry3-stack:
Part of: #2233

## In plain English
Right now the Daring AI decides to "go for the moon" based only on how many hearts it holds, not how *good* those hearts are or whether it can actually control the board. That's why real moon success is only about 5 in 100 attempts. This issue replaces the counting rule with a proper hand-quality check so Daring only commits when it has a realistic shot.

## User story
As a player facing the Daring persona, I want its moon attempts to reflect genuine hand strength (not just heart count) so that moon shots feel purposeful and the attempt/success gap narrows toward a defensible target.

## Context
The moon-attempt trigger lives in `frontend/src/game/hearts/ai.ts:207-225` (`selectCardToPlayUtility`) and a parallel copy in the pass phase at `ai.ts:113-121` (`selectCardsToPassUtility`, `moonViable`/`strongMoon`):

```ts
// ai.ts:217-223
const earlyMoon = heartsInHand >= 7 && myHasQ && heartsWon === 0 && hand.length >= 8;
const midMoon = totalHearts >= 6 && myHasQ && myPoints === totalPointsTaken && hand.length >= 5;
isMoonAttempt = earlyMoon || midMoon;
```

```ts
// ai.ts:117-121
const heartsInHand = hand.filter((c) => c.suit === "hearts").length;
const hasQSpades = hand.some(isQueenOfSpades);
const moonViable = heartsInHand >= 6 && hasQSpades;
const strongMoon = heartsInHand >= 7 && hasQSpades;
```

Both gates are pure counting rules: any 6-7 hearts qualify regardless of rank (a hand of six low hearts 2-7 scores identically to six top hearts A-K-Q-J-10-9), and Q♠ is a hard requirement even though a moon shot with strong side-suit control and no Q♠ risk can still be viable. Per #2204's instrumented rerun (n=33,325 hands), true *paired* moon success (attempt → actual 26-point completion) is **4.9%**, well under the historical ~10% target set in #1647. #1647's own investigation already names "moon-viable passing activates when Hard is dealt 5+ hearts + Q♠ but completion is rare" as the core issue — this story is the direct fix.

This is a strategy-quality change; #2159 (persona weight audit) depends on the trigger being sound before it can meaningfully tune `DARING_MOON_PLAY_WEIGHTS`.

## Acceptance criteria
[...]

Issue #2233, opened 2026-07-19
Epic: Hearts AI robustness — strategy gaps + evaluation overhaul


wcmchenry3-stack:
## In plain English
The Hearts computer players are already decent (utility-AI architecture shipped in #2022), but a July 2026 code-level gap analysis found the moon-shot trigger fires on heart *quantity* not hand *quality*, moon defense only reacts after the fact, several textbook play heuristics are missing from the utility layer, the AI can't remember what it passed or received, and — most importantly — the simulation harness that would let us safely tune any of this has broken/limited metrics and isn't wired into CI. This epic collects the 7 stories needed to close those gaps: 4 strategy stories, 2 evaluation/simulation-infrastructure stories, and 1 research spike into search-based play.

## Problem statement

**Current state:**
- Hearts CPU is a utility AI (weighted considerations + argmax + noise, no search/lookahead). Core files: `frontend/src/game/hearts/ai.ts` (`selectCardToPlayUtility` ai.ts:195-307, `selectCardsToPassUtility` ai.ts:89-178), `aiConsiderations.ts` (6 considerations + `computePWin`), `aiWeights.ts` (persona weight maps + `NOISE_RATE`: cautious 25%, schemer 10%, daring 0%), `aiInfoSet.ts` (`seenKeys` + `voidLedger`, no pass-phase memory).
- Moon attempts are Daring-only, triggered by a hardcoded heart *count* + Q♠ (`ai.ts:207-225`) that ignores heart rank and side-suit quality. Verified paired moon success is ~4.9% (#2204) vs. the ~10% historical target (#1647).
- Moon defense (`rateMoonThreat`, `aiConsiderations.ts:251-290`) is purely reactive — it only fires once one other player already holds all points taken so far (≥4 pts). No proactive stopper-retention or early-detection behavior exists.
- Several standard Hearts heuristics are absent from the utility layer: no preference for ducking with the highest safe card (verified regression candidate vs. the old rule-based AI, #1500), no "win big when forced to win" bias, no spade-flushing to force out Q♠, no endgame low-card retention.
- The information set has no memory of the passing phase — the 3 cards passed away are certain knowledge, and the 3 cards received carry inference value, but neither feeds `computePWin` or `voidLedger`.
[...]

Issue #2211, opened 2026-07-19
ci: add TypeScript type-checking to CI (123 errors / 34 files accumulated since April)


wcmchenry3-stack:
## In plain English
No CI workflow and no npm script runs `tsc` anywhere in the frontend project, so type errors have been silently accumulating unnoticed for months — the frontend currently has 123 production TypeScript errors spread across 34 files.

## Context
TypeScript type errors provide no protection today because nothing in CI checks for them. The error count (123 across 34 non-test files) has been accumulating since April 2026, confirming this has been unenforced for a long time, not a recent regression. Two of the 123 errors are a known, runtime-benign `useRef` generic issue in `DropTarget.tsx`, also noted as part of the shared drag-system test-coverage story.

## Acceptance criteria
- [ ] An `npm run typecheck` (or equivalent) script exists in `frontend/package.json` that runs `tsc --noEmit`.
- [ ] A CI workflow step runs the typecheck script on every PR and fails the build on any error.
- [ ] All 123 currently-accumulated production errors are either fixed or (for any that require larger refactors) explicitly and individually suppressed with a tracked follow-up, so the new CI check starts green.
- [ ] The known `DropTarget.tsx:31,69` `useRef` generic issue (missing `| undefined`) is fixed as part of clearing the baseline.

## Technical notes
- Baseline measured: exactly 123 errors across 34 non-test files; zero existing workflow or npm script runs `tsc` anywhere in the frontend.
- `DropTarget.tsx:31,69` — 2 of the 123 errors are a `useRef` generic missing `| undefined`; runtime-benign but should be fixed as part of establishing a clean baseline.
- Consider `tsc --noEmit` in a dedicated CI job rather than folding into an existing lint/test job, so failures are clearly attributable to type errors.

## Test coverage
- This story is itself a CI/tooling addition, not application test coverage — "test coverage" here means the CI job's own correctness (verify it fails on a deliberately introduced type error in a draft/throwaway commit before merging the final version).

## Documentation to update
- `docs/TESTING.md` — document the new typecheck step and how to run it locally.

## Out of scope
- Fixing every one of the 123 errors with maximal type-safety improvements — the goal is a clean, enforced baseline, not a broader type-safety audit.

## Dependencies
- None.

## Definition of done
CI runs `tsc --noEmit` on every PR, currently passes against a cleaned-up baseline (0 errors), and a new type error introduced in a test PR causes the CI check to fail.

[...]

PR #2247, opened 2026-07-20
test(hearts): fix sim calibration metrics and --count semantics (#2204)

### LINKED ISSUE #2204: test: fix Hearts AI calibration metrics + add ai.calibrate.test.ts
## In plain English
The Hearts AI simulator's calibration metrics are broken in ways that hide real regressions and mis-diagnose non-existent ones. Two metrics need fixing and one built-in check needs its expected direction corrected, before we can trust any Hearts AI calibration signal in CI.

## Context
This surfaced during the July 2026 full-game QA review of Hearts. Three independent defects were found in `scripts/simulate-hearts.ts`, all verified with fresh reruns:

1. **Unpaired "moon success" metric is wildly inflated.** `simulate-hearts.ts:452-497` computes "earlyMoon Success" as `all moons / narrow-trigger denominator` — an unpaired ratio. It reports 48-80% depending on run, but the true *paired* success rate (moon attempts that were actually preceded by the early/mid-moon trigger, measured against real attempts) is 4.6-4.9% (n=33,325 hands, instrumented; a fresh disjoint-seed rerun at n=300 matched). A moon-logic regression could silently hide behind this inflated metric and still pass the existing ≥15% threshold check.
2. **"Cautious vs Schemer: win rate drops" check has its expected direction backwards.** `simulate-hearts.ts:535` asserts win rate *drops* going from Cautious to Schemer persona, but the measured effect *rises* (26.3% → 28.6% in the initial run). This isn't noise: reproduced at two additional disjoint seeds, pooled across a 9,000-game batch, z ≈ 2.9, p < 0.01. The check's premise about which direction the effect should go is simply wrong, and it currently only doesn't fail because the tolerance happens to be borderline-passing by chance.
3. **`--count` semantics diverge from the Yacht simulator.** `simulate-hearts.ts:335-363` treats `--count` as an NDJSON-dump-mode row count, while `simulate-yacht.ts:254` treats it as a batch size; Hearts additionally hardcodes a 3000-game batch internally regardless of the flag. This makes the two CLIs' `--count` flag mean different things, which is confusing and error-prone when writing calibration scripts that target both.

None of this is currently caught by CI because no automated calibration test exists for Hearts AI at all — the metrics above are only visible when a human reads the simulator's console output.

## Acceptance criteria
- [ ] "earlyMoon Success" metric is repaired to compute a **paired** rate: moon attempts that occurred immediately after an actual `earlyMoon`/`midMoon` trigger activation, divided by the count of real trigger activations (not the inflated unpaired ratio)
- [ ] The "Cautious vs Schemer" built-in check's expected direction is corrected to "win rate rises" (or the check is reframed with the corrected premise), based on the pooled 9,000-game reproduction
- [ ] `--count` semantics are aligned between `simulate-hearts.ts` and `simulate-yacht.ts` (pick one convention — batch size is recommended since it's used by the more actively calibrated Yacht simulator — and document it)
- [ ] New `frontend/src/game/hearts/__tests__/ai.calibrate.test.ts` is added, using:
  - paired moon-success metrics (not the old unpaired ratio)
  - multi-seed pooled batches (not single-seed runs, to avoid the borderline-significance trap that hid defect #2 above)
  - the corrected Cautious-vs-Schemer direction
  - proposed bands based on the reproduced measurements: paired earlyMoon/midMoon success in the ~4-6% range (not 15%+); Cautious-vs-Schemer win rate expected to *rise* by roughly 1-3pp, not drop

## Technical notes
- Files: `frontend/src/game/hearts/scripts/simulate-hearts.ts` (or wherever `simulate-hearts.ts` lives relative to repo root — confirm exact path during implementation), `frontend/src/game/yacht/__tests__/ai.calibrate.test.ts` (existing Yacht pattern to mirror for `--count` alignment and test structure)
- Archived simulation data referenced during verification: 18,000-game Hearts run instrumented for paired moon-attempt/trigger tracking

### Unit tests
- [ ] `frontend/src/game/hearts/__tests__/ai.calibrate.test.ts` — new file per acceptance criteria above

## Documentation to update
- None

## Out of scope
- Rebalancing the Hearts AI's actual moon-shooting or persona weights — this issue is metrics/test-infrastructure only, not AI behavior change
- Yacht simulator changes beyond aligning `--count` semantics (see the sibling issue for Yacht simulator band recalibration)

## Dependencies
- Part of: none
- Blocked by: none
- Blocks: the sibling Yacht simulator band-recalibration issue references this one as related tooling-quality work (not a hard dependency)

## Definition of done
- [ ] All acceptance criteria met
- [ ] `ai.calibrate.test.ts` added and green in CI
- [ ] CI green (lint, type check, tests)
- [ ] PR reviewed and merged to dev

Found in the July 2026 full-game QA review (see docs/hearts-qa-report-2026-07.md on branch review/qa-2026-07); independently verified with fresh reruns at multiple disjoint seeds.

wcmchenry3-stack:
Closes #2204. QA findings HRT-1, HRT-2, HRT-3 (`docs/hearts-qa-report-2026-07.md`, branch `review/qa-2026-07`).

## What changed

**HRT-1 — unpaired moon-success metric (the 48–80% fiction).** The old "earlyMoon Success" divided *all* moon completions by a narrow trick-0-only attempt count — mismatched denominators. The sim now instruments the real trigger: `ai.ts`'s moon-attempt check is extracted as exported `detectMoonAttempt()` (pure refactor, behavior identical) and the sim counts a hand as *attempted* when that trigger actually fires for a seat, and a *paired success* only when that same seat completes the moon in that same hand. Reported as two separate metrics (attempt rate over hands; success rate over attempts), never blended.

**HRT-3 — reversed direction check.** "Cautious vs Schemer: win rate drops" was empirically backwards (pooled 9,000 games: rises ~1–3pp, z≈2.9, p<0.01). The check now expects a rise, with a −1pp tolerance so ordinary batch noise doesn't flag the correct direction.

**HRT-2 — `--count` semantics.** `--count N` now sets games-per-batch (aligned with `simulate-yacht.ts`); NDJSON dump mode moves to `--log-games N` (flag name and output format kept stable for `hearts-analysis`'s `/api/simulate`).

**New calibration gate**: `frontend/src/game/hearts/__tests__/ai.calibrate.test.ts`, mirroring the yacht gate — skipped by default, enabled with `HEARTS_SIM_FULL=<N>`, seeded. Asserts Cautious baseline win rate 20–30% and Daring paired moon bands (attempt 8–16%, paired success 2–10%). #2234 is expected to raise the success band deliberately, with sim evidence.

## Verification

- `npx jest src/game/hearts`: 7 suites / 332 tests pass; calibrate suite self-skips without the env flag.
- `HEARTS_SIM_FULL=500 npx jest …ai.calibrate.test.ts`: both calibration tests pass on live sim data.
- Full default gate (`npx tsx scripts/simulate-hearts.ts`, 6 batches × 3,000 games) — all 6 interpretation checks pass:
  - Cautious baseline 26.3% ✓
  - Cautious vs Schemer Δ+2.3pp rise, p<0.05 ✓ (direction fix confirmed)
  - Daring moon attempt rate 9.8% ✓, paired success 7.7% ✓ (vs old metric's 48–80%)
- eslint + prettier clean on touched files; `tsc` shows no new errors (the new test file inherits the same pre-existing jest-types config errors as the yacht calibrate test, tracked in #2211).

Part of the measurement-first track for epic #2233 (next: #2238 sim gate v2, which builds on this).

https://claude.ai/code/session_01UNjGcMZkrPPbXQDmpFCaUe

### DIFF
--- a/frontend/src/game/hearts/__tests__/ai.calibrate.test.ts
+++ b/frontend/src/game/hearts/__tests__/ai.calibrate.test.ts
@@ -1,0 +1,167 @@
+/**
+ * Hearts AI calibration gate (GH #2204, HRT-1/HRT-3).
+ *
+ * Full N-game validation of persona calibration bands, mirroring the Yacht
+ * gate (yacht/__tests__/ai.calibrate.test.ts). Skipped by default; enable
+ * with HEARTS_SIM_FULL=<N> (e.g. HEARTS_SIM_FULL=3000).
+ *
+ * Acceptance bands (measured on an instrumented 3,000-game / 33,325-hand run,
+ * see #2204):
+ *   Cautious baseline win rate (uniform Cautious field):    20–30%
+ *   Daring moon attempt rate (Daring seat 0 vs S/C/C):      8–16% of hands
+ *   Daring PAIRED moon success (completions ÷ activations): 2–10% of attempts
+ *
+ * The paired-success band replaces the old unpaired ratio (all completions ÷
+ * trick-0-only attempt count), which reported 48–80% against a true paired
+ * rate of ~4.9%. #2234 (rank-aware attempts) is expected to raise this band
+ * deliberately; update the numbers there, with sim evidence, when it lands.
+ */
+
+import { detectMoonAttempt, selectCardToPlay, selectCardsToPass } from "../ai";
+import {
+  commitPass,
+  createSeededRng,
+  dealGame,
+  dealNextHand,
+  playCard,
+  selectPassCard,
+  setRng,
+} from "../engine";
+import type { AiPersona, HeartsState } from "../types";
+
+const RUN = !!process.env.HEARTS_SIM_FULL;
+const N = process.env.HEARTS_SIM_FULL ? parseInt(process.env.HEARTS_SIM_FULL, 10) : 3000;
+
+const itFull = (RUN ? it : it.skip) as jest.It;
+
+type Difficulties = [AiPersona, AiPersona, AiPersona, AiPersona];
+
+// ---------------------------------------------------------------------------
+// Simulator (trimmed from scripts/simulate-hearts.ts — win + paired moon only)
+// ---------------------------------------------------------------------------
+
+interface SimResult {
+  win: boolean;
+  handsPlayed: number;
+  moonAttemptsBySeat0: number;
+  pairedMoonSuccessBySeat0: number;
+}
+
+function simulateGame(difficulties: Difficulties, seed: number): SimResult {
+  setRng(createSeededRng(seed));
+  let state: HeartsState = dealGame(difficulties[0]); // aiDifficulty field is informational
+
+  // Paired moon instrumentation (#2204): a hand counts as "attempted" for seat 0
+  // the first time the real earlyMoon/midMoon trigger (detectMoonAttempt — the
+  // same check ai.ts uses to activate DARING_MOON_PLAY_WEIGHTS) fires, and as a
+  // paired success only when that SAME hand ends with seat 0 shooting the moon.
+  let attemptedThisHand = false;
+  let moonAttempts = 0;
+  let pairedSuccesses = 0;
+
+  function collectHandEvents(s: HeartsState) {
+    for (const e of s.events ?? []) {
+      if (e.type === "moonShot" && e.shooter === 0 && attemptedThisHand) {
+        pairedSuccesses++;
+      }
+    }
+  }
+
+  while (state.phase !== "game_over") {
+    if (state.phase === "passing") {
+      for (let i = 0; i < 4; i++) {
+        const hand = [...(state.playerHands[i] ?? [])];
+        const cards = selectCardsToPass(hand, state.passDirection, difficulties[i]!, i);
+        for (const card of cards) {
+          state = selectPassCard(state, i, card);
+        }
+      }
+      state = commitPass(state);
+    } else if (state.phase === "playing") {
+      const playerIndex = state.currentPlayerIndex;
+      const hand = [...(state.playerHands[playerIndex] ?? [])];
+      if (
+        playerIndex === 0 &&
+        !attemptedThisHand &&
+        detectMoonAttempt(hand, state, playerIndex, difficulties[0]!)
+      ) {
+        attemptedThisHand = true;
+        moonAttempts++;
+      }
+      const card = selectCardToPlay(
+        hand,
+        [...state.currentTrick],
+        state,
+        playerIndex,
+        difficulties[playerIndex]!
+      );
+      state = playCard(state, playerIndex, card);
+    } else if (state.phase === "dealing") {
+      collectHandEvents(state);
+      attemptedThisHand = false;
+      state = dealNextHand(state);
+    }
+  }
+  collectHandEvents(state);
+
+  return {
+    win: state.winnerIndex === 0,
+    handsPlayed: state.handNumber,
+    moonAttemptsBySeat0: moonAttempts,
+    pairedMoonSuccessBySeat0: pairedSuccesses,
+  };
+}
+
+interface BatchResult {
+  winRate: number;
+  totalHands: number;
+  moonAttemptRate: number;
+  pairedMoonSuccessRate: number;
+}
+
+function runBatch(difficulties: Difficulties, n: number, seedOffset: number): BatchResult {
+  let wins = 0;
+  let hands = 0;
+  let attempts = 0;
+  let successes = 0;
+  for (let i = 0; i < n; i++) {
+    const r = simulateGame(difficulties, seedOffset + i);
+    if (r.win) wins++;
+    hands += r.handsPlayed;
+    attempts += r.moonAttemptsBySeat0;
+    successes += r.pairedMoonSuccessBySeat0;
+  }
+  return {
+    winRate: wins / n,
+    totalHands: hands,
+    moonAttemptRate: hands > 0 ? attempts / hands : 0,
+    pairedMoonSuccessRate: attempts > 0 ? successes / attempts : 0,
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+afterEach(() => setRng(Math.random));
+
+describe("Hearts calibration — persona bands (#2204)", () => {
+  itFull("Cautious baseline: seat 0 win rate 20–30% in a uniform Cautious field", () => {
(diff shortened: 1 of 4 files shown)


wcmchenry3-stack:
# Review: PR #2247 — test(hearts): fix sim calibration metrics and `--count` semantics

## Overview

Three QA fixes from the July review: (1) replaces the meaningless "earlyMoon Success" metric (mismatched numerator/denominator, reporting 48–80% vs a true ~4.9%) with paired attempt→outcome tracking, instrumented via a newly extracted `detectMoonAttempt()` so the sim counts the *real* AI trigger; (2) reverses the empirically backwards Cautious-vs-Schemer direction check; (3) swaps `--count` to mean games-per-batch (matching `simulate-yacht.ts`) and moves NDJSON dump mode to `--log-games`. Adds a skip-by-default calibration gate mirroring the yacht one.

## What checks out ✓

- **`detectMoonAttempt()` extraction is a verbatim pure refactor** — diffed against the removed inline block character-for-character; identical logic, including the documented intentional `totalPointsTaken === 0` trivial-fire. And since `selectCardToPlay` delegates unconditionally to `selectCardToPlayUtility` (`ai.ts:359-366`), the instrumented trigger genuinely *is* the activation path — the stated goal of the extraction holds.
- **The pairing logic is sound.** `dealNextHand` resets `events: []` (`engine.ts:147`), so `collectHandEvents` at the dealing boundary sees exactly one hand's events, and `attemptedThisHand` is reset in the same branch. `handNumber` starts at 1 and increments per deal, so `handsPlayed: state.handNumber` at game-over is correct.
- The methodology fix itself (attempt rate over hands, success rate over attempts, never blended) is the right correction, and the test bands match the PR's measured data with sensible tolerance. `afterEach(() => setRng(Math.random))` correctly restores global RNG.

## 🔴 Blocking: this breaks `hearts-analysis`'s `/api/simulate`

The PR body and the code comment both claim `--log-games` is "used by hearts-analysis/main.py's /api/simulate endpoint, so the flag name … must stay stable." **That's not what main.py does.** On `origin/dev` (the PR base), `hearts-analysis/main.py:116-120` invokes:

```python
cmd = ["npx", "tsx", "scripts/simulate-hearts.ts", "--count", str(safe_count), ...]
```

It uses `--count` — which *was* the primary NDJSON dump flag (`--log-games` was the deprecated alias, per the removed comment). The PR changes only 3 files and doesn't touch main.py, so after merge, `/api/simulate` will:

1. Run **aggregate-stats mode** — 6 batches × `safe_count` games (up to 3,000 games under main.py's 120 s subprocess timeout, which it will likely blow), and
2. Write human-readable stats tables into `games.json`, which `_reload_games()` expects to be NDJSON — so the endpoint silently loads 0 games or errors.

**Fix:** update `main.py:118` to pass `--log-games` in this PR (it's a one-token change), and correct the comment in `simulate-hearts.ts` — as written it documents the intended end-state as if it were the current one, which is exactly how this slipped through.

Side observation while I was in there (pre-existing, not this PR's fault, but worth a follow-up issue): main.py sends `--difficulties medium,medium,...` (easy/medium/hard) while the script validates cautious/schemer/daring — `parseDifficulties` returns null and the API's difficulty parameter is silently ignored.

## Minor issues

- **`--count abc` silently runs with the default 3000** — `parseCount` returns `null` on non-numeric input, so the `GAMES_PER_BATCH < 1` guard never sees it. A typo'd flag value produces a full default run instead of an error. Low stakes for a dev script, but the guard implies stricter validation than exists.
- **`HEARTS_SIM_FULL=true` (non-numeric) → `N = NaN`** — the loop runs zero games and the assertion fails with a confusing `expect(NaN)` message rather than "invalid N". A `Number.isFinite` guard with a clear message would save the next person a puzzled minute. (If the yacht gate shares this, fix both or neither.)
- **The calibrate test duplicates the sim loop from `simulate-hearts.ts`** rather than sharing it. The header honestly says "trimmed from," and importing from `scripts/` into a jest suite is awkward — but this is the same drift-by-reimplementation risk the PR itself fixes for `detectMoonAttempt`. Fine to accept now; consider extracting a shared harness in #2238 (sim gate v2), which seems like the natural home.
- **Cosmetic:** the HRT-3 check prints "✓ … win rate rises" for any Δ > −1pp, so a small *negative* delta still displays as a passing "rises" line. The rationale comment justifies the tolerance well; just consider wording the label "does not drop (expected ~1–3pp rise)" so the printed claim matches what was actually asserted.

## Conventions, tests, security

- Style matches the surrounding code well — comment density, issue-number anchoring, and the yacht-gate mirroring are all consistent with house patterns. Branch/PR flow follows the git standard.
- Coverage: `detectMoonAttempt` is exercised on every Daring play in the existing 332-test suite, and the gate ran green at `HEARTS_SIM_FULL=500` plus a full 6-batch sim per the PR body. Skip-by-default gating matches the established yacht precedent.
- No security concerns — no new inputs beyond dev-script flags, and `parseInt` on argv/env is benign here.

## Verdict

The metrics/statistics work is careful and verifiably correct — the extraction, pairing logic, and direction fix all held up under inspection. But **request changes** for the `/api/simulate` regression: the PR's own compatibility claim is contradicted by `main.py:118` on the base branch, and the one-line main.py update belongs in this PR so the flag rename and its only consumer move together.

https://claude.ai/code/session_01NueMZ3hKAvV4BCcE2n2KCY

wcmchenry3-stack:
Pushed `5deb8493` to address the review:

- **Blocking fix:** `hearts-analysis/main.py:118` now passes `--log-games` instead of `--count`, so `/api/simulate` still triggers NDJSON dump mode after the flag swap. Verified `--log-games` still emits one NDJSON line per game and `--count` still runs aggregate-stats mode with the new games-per-batch semantics.
- **Cosmetic fix:** the HRT-3 Interpretation line now reads "win rate does not drop" instead of "win rate rises", matching the actual `csDelta > -0.01` tolerance being checked (a small negative delta was previously mislabeled as a rise).

Left as-is, deliberately:
- The `--count`/`HEARTS_SIM_FULL` non-numeric-input silent-default behavior — `simulate-yacht.ts` and its calibrate test have the identical pattern (checked both). Fixing only the hearts side would break parity with the script this PR explicitly mirrors; happy to file a follow-up covering both scripts together if that's wanted.
- The pre-existing `main.py` difficulty-vocabulary mismatch (`easy/medium/hard` sent vs `cautious/schemer/daring` expected) — unrelated to this PR's diff, flagged for a separate issue.

All 332 hearts jest tests still pass; prettier clean on the touched TS file.

